PR #3, opened 2018-08-17
Fix variable name shadowing.

thvasilo:
Project would not compile under GCC because of variable name shadowing (local int flags vs. enum)


AlexanderSaydakov:
I rolled back my name change of the enum.
Thank you.

PR #13, opened 2018-11-15
Merge HLL due to structural changes

jmalkin:
HLL isn't fully tested as much as I'd like, but some of the structural changes to the make process (we do need to look into cmake at some point) were easier to test with multiple modules to compile. So before things diverge further, trying to merge.

The memory leak reported in the newsgroup has _not_ yet been resolved. Bureaucracy will delay my effort to address that for a bit longer.
https://groups.google.com/forum/#!topic/sketches-user/zwIERlunvn4


jmalkin:
Ugh, had NDEBUG enabled. Gotta remove that!

PR #14, opened 2018-11-21
switch hll to return unique_ptr in public API, some makefile refinements

jmalkin:


jmalkin:
This PR also fixes a few incorrect delete calls (needed to be delete [] instead) as pointed out in the usergroup: https://groups.google.com/forum/#!topic/sketches-user/zwIERlunvn4

PR #19, opened 2019-01-02
Needed #include <cmath> to fix compile error on "std::isnan(value)".

leerho:
I also get two compiler warnings:

moving a local object in a return statement prevents copy elision [-Wpessimizing-move]	cpc_sketch.hpp	/sketches-core-cpp/cpc/include	line 385	C/C++ Problem

moving a local object in a return statement prevents copy elision [-Wpessimizing-move]	cpc_sketch.hpp	/sketches-core-cpp/cpc/include	line 483	C/C++ Problem

Please check.

### DIFF
--- a/cpc/include/cpc_sketch.hpp
+++ b/cpc/include/cpc_sketch.hpp
@@ -10,6 +10,7 @@
 #include <memory>
 #include <functional>
 #include <stdexcept>
+#include <cmath>
 
 #include "fm85.h"
 #include "fm85Compression.h"


leerho:
But clang is what the makefile specifies: config.mk line 5.

jmalkin:
Yes, and you can fairly trivially modify it to use g++ instead. We'll ultimately want to have a configure script that auto-detects the default compiler, I suspect.

AlexanderSaydakov:
This seems to be outdated. I think we are good in master now

PR #20, opened 2019-01-22
Python

jmalkin:
Python3 wrappers for HLL and KLL. No Makefile yet since there are enough config options it might need some sort of configure script.


jmalkin:
Adds python wrappers, but also fixes some HLL bugs uncovered in the process of testing the wrapper

PR #23, opened 2019-02-04
rewrote checking to avoid C++17 feature

AlexanderSaydakov:


AlexanderSaydakov:
Turns out that enable_if_t is a C++14 feature. Let me see if I can do this in C++11.

jmalkin:
Ok, clang is happy now, too. Time to merge!

PR #39, opened 2019-06-24
Placement new

AlexanderSaydakov:


jmalkin:
Not obvious what gcc doesn't like about those tests, but removing those means we have no clue if any of this code is internally consistent or consistent with java's handling of inputs. So that's very bad.

The placement new changes are good.

AlexanderSaydakov:
We need to get back to these tests to see what we can restore, and what we should remove.

jmalkin:
The HLL unit tests were only using the public API. That the code didn't compile under gcc means that there's an actual issue. It shows up with the python wrapper, too.

AlexanderSaydakov:
@jmalkin I see you restored the branch. Do we need it?

jmalkin:
I re-opened since once of the tests that was disabled really needed to be there, but that wasn't really the right way to handle the issue after the branch was already merged. And subsequent work (immediately after) fixed the underlying issue,

PR #41, opened 2019-06-26
change another location to placement new, restore unit tests

jmalkin:
I don't have a gcc-compatible cppunit, but I think this should fix the private access issue that broke the unit tests (which were using only the public API!)


AlexanderSaydakov:
very cool

PR #43, opened 2019-06-26
compatibility changes for gcc 4.x on old RHEL

jmalkin:


jmalkin:
The makefile changes will break things on the mac, but i'm considering the make stuff deprecated. Keeping it in case people on older systems refuse to use a more modern build system than hand-generated Makefiles.

PR #52, opened 2019-08-02
change HLL public API from CamelCase to lower_with_underscore

jmalkin:
I think I got the public portions of hll.hpp changed. This can be extended to that whole file if we think it's warranted.


jmalkin:
Actually, let's hold off on this a bit longer. I think I can do more fairly quickly.

jmalkin:
Now I think this is ready for review. Renaming HllSketchImpl will take more work with less obvious short-term benefit.

AlexanderSaydakov:
looks good to me

PR #53, opened 2019-08-06
Fix cmake install

gabm:
some header files got removed in #50 but the install command still has those files listed...

on a side-note: why have headers like "HllSketch.hpp" been removed?


jmalkin:
Ah, forgot to remove those references to the files.

When initially writing HLL it mirrored the Java version exactly. But we ultimately decided that, since C++ doesn't limit you to a single class name per file like Java, that it'd be a nicer interface to define both the sketch and union in a single header, and have that one file contain the entire public API. HllSketch.hpp and HllUnion.hpp were entirely commented out, so as part of code cleanup we (read: @AlexanderSaydakov) remembered to remove them. Because I kept forgetting to do so.

Just FYI, I have a branch converting HLL from camelCase to lower_with_underscores (still in progress). Eventually all the code, but we're trying to get a consistent interface style across all our classes before the first formal release.

gabm:
Thanks for the heads-up..

PR #55, opened 2019-08-06
remove default construction in deserialize

AlexanderSaydakov:


jmalkin:
Deserialization already initializes the items so empty initialization in this case was useless?

AlexanderSaydakov:
Yes, a leftover from the old way. Forgot to remove as a part of the previous change.

PR #65, opened 2019-09-19
Py test

jmalkin:
adds unit test coverage for python wrappers, mostly with tutorial-like example tests


jmalkin:
I have had one of the kll tests fail on rare occasions.  Forgot to note which one so I could increase the error margin a bit more.

AlexanderSaydakov:
looks good to me

PR #71, opened 2019-11-21
theta trim method

AlexanderSaydakov:


jmalkin:
I think this looks good. Is it ready to merge?

PR #81, opened 2020-01-22
not building shared library anymore

AlexanderSaydakov:


AlexanderSaydakov:
We completed our planned move to header-only library. Therefore the only code to compile is unit tests.

jmalkin:
Since we only build for unit testing, is the (more fragile) cmake alternative even needed?  Should we just remove the Makefile option entirely?

AlexanderSaydakov:
I agree. Let us reduce unnecessary complexity.

PR #85, opened 2020-02-13
HLL union performance improvement

AlexanderSaydakov:


jmalkin:
did a cursory readthrough that seemed generally ok. can do a more look tomorrow

hll/include/HllArray-internal.hpp
```diff
@@ -26,11 +26,11 @@
 #include "CubicInterpolation.hpp"
 #include "CompositeInterpolationXTable.hpp"
 #include "CouponList.hpp"
-
 #include <cstring>
 #include <cmath>
 #include <stdexcept>
 #include <string>
+#include "../../common/include/inv_pow2_table.hpp"
```
jmalkin: i think convention is to put project-internal files first?

hll/include/HllUnion-internal.hpp
```diff
@@ -301,169 +309,41 @@ inline HllSketchImpl<A>* hll_union_alloc<A>::leak_free_coupon_update(HllSketchIm
...
-      break;
-    }
-    case 4: { //src: LIST, gadget: SET
-      pair_iterator_with_deleter<A> srcItr = src_impl->getIterator(); //LIST
-      while (srcItr->nextValid()) {
-        dstImpl = leak_free_coupon_update(dstImpl, srcItr->getPair()); //assignment required
+void hll_union_alloc<A>::union_impl(const hll_sketch_alloc<A>& sketch, const int lg_max_k) {
+  const HllSketchImpl<A>* src_impl = sketch.sketch_impl; //default
+  HllSketchImpl<A>* dst_impl = gadget.sketch_impl; //default
+  if (src_impl->getCurMode() == LIST or src_impl->getCurMode() == SET) {
```
jmalkin: consider just using the standard && and || -- i guess there's potential confusion between a double reference && vs logical and, but those should be trivially obvious from the context

AlexanderSaydakov: agree. perhaps we should replace everywhere as a separate task

hll/include/AuxHashMap.hpp
```diff
@@ -20,9 +20,11 @@
 #ifndef _AUXHASHMAP_HPP_
 #define _AUXHASHMAP_HPP_
 
-#include "PairIterator.hpp"
-
+#include <iostream>
```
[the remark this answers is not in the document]
AlexanderSaydakov: because of deserialize(std::istream...) declaration

hll/include/CouponList-internal.hpp
```diff
@@ -23,11 +23,7 @@
 #include "CouponList.hpp"
 #include "CubicInterpolation.hpp"
 #include "HllUtil.hpp"
-#include "IntArrayPairIterator.hpp"
 
-#include <iostream>
-#include <cstring>
-#include <cmath>
```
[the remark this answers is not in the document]
AlexanderSaydakov: Mistake, which does not show up in my environment. I will put it back

hll/include/HllArray-internal.hpp
AlexanderSaydakov: I am not aware of this convention. Perhaps we should discuss.

PR #87, opened 2020-02-17
Varopt Sampling

jmalkin:
I wish I could have broken this up more, but I actually ended up finding a number of issues only when I did the python integration. So now I'm torn on whether it'd be better to do PRs of "working-but-incomplete" sketches vs a giant PR like this.

Varopt only, no straight reservoir sampling (varopt will work for that but with non-trivial additional overhead). Tested, and with python wrapper and test/example code.


sampling/include/var_opt_sketch.hpp
```diff
@@ -0,0 +1,249 @@
...
+template <typename T, typename S, typename A> class var_opt_union; // forward declaration
+
+template <typename T, typename S = serde<T>, typename A = std::allocator<T>>
+class var_opt_sketch {
+
+  public:
+    static const resize_factor DEFAULT_RESIZE_FACTOR = X8;
+    static const uint32_t MAX_K = ((uint32_t) 1 << 31) - 2;
+
+    var_opt_sketch(uint32_t k, resize_factor rf = DEFAULT_RESIZE_FACTOR);
```
AlexanderSaydakov: I think this should be marked as explicit to avoid surprises

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+  }
+}
+
+/* In the "light" case the new item has weight <= old_tau, so
+   would appear to the right of the R items in a hypothetical reverse-sorted
+   list. It is easy to prove that it is light enough to be part of this
+   round's downsampling */
+template<typename T, typename S, typename A>
+void var_opt_sketch<T,S,A>::update_light(const T& item, double weight, bool mark) {
+  assert(r_ >= 1);
```
AlexanderSaydakov: I believe we had an agreement to throw std::logic_error or something

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+
+    assert(old_final_r_idx == k_);
+    swap_values(old_final_r_idx, old_gap_idx);
+
+    // now we pull an item out of H; any item is ok, but if we grab the rightmost and then
+    // reduce h_, the heap invariant will be preserved (and the gap will be restored), plus
+    // the push() of the item that will probably happen later will be cheap.
+
+    uint32_t pulled_idx = h_ - 1;
+    T pulled_item = data_[pulled_idx];
```
AlexanderSaydakov: is this intended to be a copy?

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+    for (int i = 0; i < prev_size; ++i) {
+      A().construct(&tmp_data[i], std::move(data_[i]));
+      A().destroy(data_ + i);
+      tmp_weights[i] = weights_[i];
+    }
+
+    A().deallocate(data_, prev_size);
+    AllocDouble().deallocate(weights_, prev_size);
+
+    data_ = std::move(tmp_data);
```
AlexanderSaydakov: no need to move here, it is a pointer

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+  const int last_slot = h_ - 1;
+  assert(slot_in <= last_slot);
+
+  int slot = slot_in;
+  int child = (2 * slot_in) + 1; // might be invalid, need to check
+
+  while (child <= last_slot) {
+    int child2 = child + 1; // might also be invalid
+    if ((child2 <= last_slot) && (weights_[child2] < weights_[child])) {
+      // siwtch to otehr vhild if it's both valid and smaller
```
AlexanderSaydakov: typo

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+
+  if (is_marked(h_)) {
+    --num_marks_in_h_;
+  }
+}
+
+
+template<typename T, typename S, typename A>
+void var_opt_sketch<T,S,A>::swap_values(int src, int dst) {
+  T item = std::move(data_[src]);
```
AlexanderSaydakov: I would suggest std::swap here

sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -0,0 +1,1591 @@
...
+
+template<typename T, typename S, typename A>
+var_opt_sketch<T,S,A>::const_iterator::const_iterator(const var_opt_sketch<T,S,A>& sk, bool is_end) :
+  sk_(&sk),
+  cum_r_weight_(0.0),
+  r_item_wt_(sk.get_tau()),
+  weight_correction_(false)
+{
+  // index logic easier to read if not inline
+  final_idx_ = (sk.r_ > 0 ? sk.h_ + sk.r_ + 1 : sk.h_);
```
AlexanderSaydakov: const ?

sampling/include/var_opt_union.hpp
```diff
@@ -0,0 +1,171 @@
...
+  var_opt_union(var_opt_union&& other) noexcept;
+  static var_opt_union deserialize(std::istream& is);
+  static var_opt_union deserialize(const void* bytes, size_t size);
+    
+  ~var_opt_union();
+
+  var_opt_union& operator=(const var_opt_union& other);
+  var_opt_union& operator=(var_opt_union&& other);
+
+  void update(var_opt_sketch<T,S,A>& sk);
```
AlexanderSaydakov: const input

sampling/include/var_opt_union_impl.hpp
```diff
@@ -0,0 +1,609 @@
...
+  typedef typename std::allocator_traits<A>::template rebind_alloc<double> AllocDouble;
+  double* wts = AllocDouble().allocate(result_k + 1);
+  T* data     = A().allocate(result_k + 1);
+    
+  // insert R region items, ignoring weights
+  // Currently (May 2017) this next block is unreachable; this coercer is used only in the
+  // pseudo-exact case in which case there are no items natively in R, only marked items in H
+  // that will be moved into R as part of the coercion process.
+  // Addedndum (Jan 2020): Cleanup at end of method assumes R count is 0
+  size_t final_idx = gadget_.get_num_samples();
```
AlexanderSaydakov: const ?

sampling/include/var_opt_sketch_impl.hpp
jmalkin: from (admittedly limited) conversations with people elsewhere, i suspect we're likely going to need to remove exceptions to make the library be considered more production-worthy

sampling/include/var_opt_sketch_impl.hpp
jmalkin: nope, that should be a move

sampling/include/var_opt_sketch_impl.hpp
jmalkin: huh, i thought there was some reason i didn't use it but i can't see why at this point.

sampling/include/var_opt_sketch.hpp
jmalkin: right, i think it slipped through since I originally didn't have a default value

sampling/include/var_opt_union_impl.hpp
jmalkin: this one matters least of all in the entire code since that block isn't executed :)

sampling/include/var_opt_sketch_impl.hpp
AlexanderSaydakov: Maybe, but asserts, if enabled, lead to exit with no way to recover, which is not production-worthy either. Perhaps, we need to bring this up on the dev list. I don't mind leaving this as it is for now.

sampling/include/var_opt_sketch_impl.hpp
jmalkin: had to rearrange a bit but made it const.. the more complex constructor now conditions on use_r_region both to initialize final_idx_ upon declaration and in the body (where inlining the logic got annoying to read) but i guess that's ok

jmalkin:
addressed everything bu the asserts.

i think std::allocator::construct() is dfeprecated in favor of the one in allocator_traits so I'm tempted to replace it with placement new everywhere, except that then you can't override with a custom constructor.

AlexanderSaydakov:
looks good to me

PR #88, opened 2020-02-17
KLL moving merge

AlexanderSaydakov:


kll/test/kll_sketch_custom_type_test.cpp
```diff
@@ -150,6 +151,17 @@ class kll_sketch_custom_type_test: public CppUnit::TestFixture {
     CPPUNIT_ASSERT_EQUAL(sketch1.get_rank(n / 2), sketch2.get_rank(n / 2));
   }
 
+  void moving_merge() {
+    kll_test_type_sketch sketch1(8);
+    for (int i = 0; i < 10; i++) sketch1.update(i);
+    kll_test_type_sketch sketch2(8);
+    sketch2.update(10);
+    sketch2.merge(std::move(sketch1));
```
jmalkin: this test seems like a good example of where having proper code coverage measures would be useful

PR #98, opened 2020-03-03
frequent items weight type as a template parameter

AlexanderSaydakov:
This will allow changing the type for weights from uint64_t. For instance, something smaller such as uint32_t or float.

### DIFF
--- a/fi/include/frequent_items_sketch_impl.hpp
+++ b/fi/include/frequent_items_sketch_impl.hpp
@@ -470,7 +470,7 @@ void frequent_items_sketch<T, W, H, E, S, A>::check_weight(WW weight) {
   if (std::isnan(weight)) {
     throw std::invalid_argument("weight must be a valid number");
   }
-  if (weight > std::numeric_limits<WW>::max()) {
+  if (std::isinf(weight)) {
     throw std::invalid_argument("weight must be finite");
   }
 }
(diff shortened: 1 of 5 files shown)


fi/include/frequent_items_sketch_impl.hpp
```diff
@@ -440,6 +445,19 @@ void frequent_items_sketch<T, H, E, S, A>::to_stream(std::ostream& os, bool prin
   }
 }
 
+template<typename T, typename W, typename H, typename E, typename S, typename A>
+template<typename WW, typename std::enable_if<std::is_signed<WW>::value, int>::type>
```
jmalkin: i believe floats and doubles need additional checks for NaN and infinity

fi/include/frequent_items_sketch_impl.hpp
```diff
@@ -445,19 +446,35 @@ void frequent_items_sketch<T, W, H, E, S, A>::to_stream(std::ostream& os, bool p
...
+template<typename T, typename W, typename H, typename E, typename S, typename A>
+template<typename WW, typename std::enable_if<std::is_floating_point<WW>::value, int>::type>
+void frequent_items_sketch<T, W, H, E, S, A>::check_weight(WW weight) {
+  if (weight < 0) {
+    throw std::invalid_argument("weight must be non-negative");
+  }
+  if (std::isnan(weight)) {
+    throw std::invalid_argument("weight must be a valid number");
+  }
+  if (weight > std::numeric_limits<WW>::max()) {
```
jmalkin: there's actually std::isinf() if you want to be more direct but i'm ok with this

PR #106, opened 2020-04-13
Catch2 tests

jmalkin:
remove dependence on external cppunit library by moving to header-only catch2 testing framework.

this _should_ help make CI easier to set up.


jmalkin:
last commit was that this is still WIP, but it seems i can't add a workflow action to an existing branch. i guess i'll try specifying this branch name and see if that works (and if so kill this PR for now)

jmalkin:
i think i figured this out.. closing for now..

PR #109, opened 2020-04-16
API and docs

AlexanderSaydakov:


theta/include/theta_sketch.hpp
```diff
@@ -53,36 +53,124 @@ class theta_sketch_alloc {
...
+   * It is an uninitialized space of a given size.
+   * This header is used in Datasketches PostgreSQL extension.
+   * @param header_size_bytes space to reserve in front of the sketch
+   */
   virtual vector_bytes serialize(unsigned header_size_bytes = 0) const = 0;
 
+  // This is a convenience alias for users
+  // The type returned by the following deserialize methods
+  // It is not possible to return instances of an abstract type, so this has to be a pointer
   typedef std::unique_ptr<theta_sketch_alloc<A>, std::function<void(theta_sketch_alloc<A>*)>> unique_ptr;
```
jmalkin, line 142: hadn't looked at this closely before.

is this too generic a name? we can only have 1 sketch in the library use it, otherwise another app that uses multiple sketch types could run into issues.

jmalkin: nevermind, this is a typedef within the class so it's fine.

PR #111, opened 2020-04-21
fix memory leak in theta_union::get_result()

jmalkin:
if num_keys == 0, we were passing nullptr to the returned sketch, but the keys array was never freed.


jmalkin:
That "failure" is ok. It was a python kll test, but my change didn't touch kll at all -- that test just occasionally fails.  Not sure how wide an error margin we should give it so that's not an issue.

PR #125, opened 2020-05-07
use vector to simplify and improve safety

AlexanderSaydakov:
sketch and union update speed is the same, sketch uses 16 more bytes of memory

### DIFF
--- a/theta/include/theta_a_not_b_impl.hpp
+++ b/theta/include/theta_a_not_b_impl.hpp
@@ -43,34 +43,32 @@ compact_theta_sketch_alloc<A> theta_a_not_b_alloc<A>::compute(const theta_sketch
   if (a.get_num_retained() == 0 || b.is_empty()) return compact_theta_sketch_alloc<A>(a, ordered);
 
   const uint64_t theta = std::min(a.get_theta64(), b.get_theta64());
-  uint64_t* keys = nullptr;
+  vector_u64<A> keys;
   uint32_t keys_size = 0;
   uint32_t count = 0;
   bool is_empty = a.is_empty();
 
   if (b.get_num_retained() == 0) {
     for (auto key: a) if (key < theta) ++count;
-    keys_size = count;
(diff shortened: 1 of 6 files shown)


theta/include/theta_a_not_b_impl.hpp
```diff
@@ -79,28 +77,22 @@ compact_theta_sketch_alloc<A> theta_a_not_b_alloc<A>::compute(const theta_sketch
     // scan A lookup B
     for (auto key: a) {
       if (key < theta) {
-        if (!update_theta_sketch_alloc<A>::hash_search(key, b_hash_table, lg_size)) keys[count++] = key;
+        if (!update_theta_sketch_alloc<A>::hash_search(key, b_hash_table.data(), lg_size)) keys[count++] = key;
```
jmalkin, line 80: Is hash_search() ever used on a raw pointer after this refactor? If no, we should just update its signature to take a vector.

theta/include/theta_intersection_impl.hpp
```diff
@@ -123,23 +60,22 @@ void theta_intersection_alloc<A>::update(const theta_sketch_alloc<A>& sketch) {
   if (!is_valid_) { // first update, clone incoming sketch
     is_valid_ = true;
     lg_size_ = lg_size_from_count(sketch.get_num_retained(), update_theta_sketch_alloc<A>::REBUILD_THRESHOLD);
-    keys_ = AllocU64().allocate(1 << lg_size_);
-    std::fill(keys_, &keys_[1 << lg_size_], 0);
+    keys_.resize(1 << lg_size_, 0);
     for (auto key: sketch) {
-      if (!update_theta_sketch_alloc<A>::hash_search_or_insert(key, keys_, lg_size_)) {
+      if (!update_theta_sketch_alloc<A>::hash_search_or_insert(key, keys_.data(), lg_size_)) {
```
jmalkin, line 65: same comment as with hash_search() above

PR #134, opened 2020-05-11
kll: check for NaN on update with floating point types

### LINKED ISSUE #133: python: kll_floats_sketch happily ingests numpy.NaN
From the dev list: https://lists.apache.org/thread.html/rd35d7f43b54b6c818ad260304e4369d6fa71b1cd552e4836d66eda28%40%3Cdev.datasketches.apache.org%3E

Easy to reproduce:
```
sk = kll_floats_sketch(10)
sk.update(numpy.nan)
sk.get_min_value()
```

Expected behavior is to ignore NaNs.

jmalkin:
Should fix #133

### DIFF
--- a/kll/include/kll_sketch.hpp
+++ b/kll/include/kll_sketch.hpp
@@ -505,17 +505,28 @@ class kll_sketch {
     static void check_serial_version(uint8_t serial_version);
     static void check_family_id(uint8_t family_id);
 
-    // implementation for floating point types
+    // implementations for floating point types
     template<typename TT = T, typename std::enable_if<std::is_floating_point<TT>::value, int>::type = 0>
     static TT get_invalid_value() {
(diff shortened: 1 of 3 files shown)


leerho:
A very simple test for NaN of floating point types could be:

```
template<typename T, typename C, typename S, typename A>
void kll_sketch<T, C, S, A>::update(const T& value) {
  if ( value != value ) { return; }
  ...
}
```

This should work for all types even non-numeric types.  Only NaNs are not equal to themselves.
This should eliminate the overhead of testing whether it is a floating-point type and the separation of floating versus non-floating point type methods.

If you look into the source of the built-in libraries that implement isNaN(value) methods, the above is exactly how it is done.

jmalkin:
The template metaprogramming is applied at compile time. This way we only eat the runtime cost of that comparison if the type requires it.

PR #137, opened 2020-05-13
Istream checks

jmalkin:
Check istream.good() flag before returning from deserialize(). Throw if not good.

This won't work if a serde uses the << operator on a string since that sets the eof flag if it reaches the stream's end. I don't think that's a viable way to ingest a set of items anyway, although I could be mistaken.


jmalkin:
This was built off the to_string branch since I wasn't sure if I'd end up conflicting with anything in there.  I don't think it actually did? But should review the to_string branch first.

PR #143, opened 2020-05-15
Added Numpy compatibility for KLL sketches

mdhimes:
Summary:
- new (wrapper) class called kll_sketches.  Creates `d` instances of kll_sketch objects.  This class is similar to the existing Python wrapper for kll_sketch, except that multiple sketches are allowed and they are compatible with Numpy.  The sketches are managed in C++ rather than Python, which makes things much faster when using many sketches in parallel.
- new wrapper methods for updating, serializing, etc. with documentation.  Some minor naming differences from the old kll_sketch wrapper (e.g., get_min_value --> get_min_values).  They accept values, lists, or Numpy arrays.  Some can select which sketch(es) to perform the action for (e.g., only get the CDF for sketch 3).

All of the additions are located in python/src/kll_wrapper.cpp.

Currently, merge is not implemented as I am unsure what the best approach to that is (e.g., once 2 sketches in the object are merged, do we delete the sketch was was merged into the other? Should this work with the Python kll_sketch object?).  This is quick to implement once this is decided.

Some considerations:
- The new class name is quite similar to the existing kll_sketch class, and this may be a bit confusing.  Perhaps a name like python_kll_sketches would better differentiate them?  I'm pretty sure that the vast majority of users of this new class would be using it via Python.
- Currently, the way it is set up requires users to use Numpy, since it returns Numpy arrays.  While I find it hard to believe someone might use a numerical algorithm in Python but not use Numpy, I'm sure this will happen, so it'd be a good idea to rework the code to allow non-Numpy users to still use it.  Perhaps the return data type can be based on the passed in data type (e.g., return a list if a list was passed in).  Though this would still require the numpy.h header file dependency, it would remove the need for the kll_sketch wrapper and simplify the Python interface to be handled only through the new class.

Let me know if there are any questions, or if there are any features desired that are not currently implemented.


jmalkin:
I wonder if we should go into a separate branch until we can get things cleaned up more? Anyway, let's at least discuss things for now.

I'm not yet sure how we'd be able to use a single interface for all kll sketches, especially not without  a performance overhead. C++ templates are processed at compile time, where they're instantiated with a concrete type. This lets them use information about that type in ways that a language like Java can't, with the cost of only being able to instantiate whatever was used during compilation. This is why there are those bind calls at the end. It also means there's doesn't seem to be a way to remove the need for a wrapper class.

For the sampling class, it does use an input type of py::object, and those can be heterogeneous types, but I don't need to do any sort of comparison between them. For kll, on the other hand, we need to ensure that all the inputs are meaningfully comparable. You can define such a function -- the sampling class can accept a predicate passed in from python -- but that's not something we'd want to put in the main critical path for updates.

So I'm not sure what the performance overhead would be of dereferencing the vector on every update, even for a 1-dimensional input, or of how we'd enforce typing with a more generic input. 

It should be easy enough to overload update with py::array_t, py::list, etc., so that part isn't a concern. But they will all need to be predefined.

jmalkin:
I can't resolve the conflicts without committing into your own repo's master. Can I ask you to re-create this PR against the python_vector_kll branch?

mdhimes:
I changed the base of the PR to the python_vector_kll branch.  Let me know if you need me to do anything else.

jmalkin:
I realized I didn't create the branch from where you forked, so I think I need to ask you to do the conflict resolution.  But it's pretty simple in this case.

kll_sketch.hpp:
- keep <sstream> and remove the assert header (we removed the asserts from code)

kll_sketch_impl.hpp
- just keep the new method it has marked as a conflict

kll_wrapper.cpp
- keep your entire contribution, obviously

Then I can do some additional cleanup (we removed to_stream() and standardized on to_string() in all sketches, for instance).

mdhimes:
Alright, I've merged in the changes.

jmalkin:
Great, now we can discuss other changes within the repo and anyone can propose commits. Thanks!

Issue #149, opened 2020-05-28
Cannot deserialize Theta EmptyCompactSketch from Java


Lordshinjo:
In Java, an empty Theta compact sketch is represented by the singleton `EmptyCompactSketch`, which has a seed hash of 0.

It is represented by the constant byte array `[1, 3, 3, 0, 0, 0x1E, 0, 0]`, as defined [here](https://github.com/apache/incubator-datasketches-java/blob/1.3.0-incubating/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java#L38).
Deserializing this byte array from Java returns `EmptyCompactSketch` as expected (snippet using [Ammonite](https://ammonite.io/)):
```
@ import $ivy.`org.apache.datasketches:datasketches-java:1.3.0-incubating`, org.apache.datasketches.theta._, org.apache.datasketches.memory._
import $ivy.$                                                           , org.apache.datasketches.theta._, org.apache.datasketches.memory._

@ val emptySketch = (new UpdateSketchBuilder).build.compact
emptySketch: CompactSketch =
### EmptyCompactSketch SUMMARY:
   Estimate                : 0.0
   Upper Bound, 95% conf   : 0.0
   Lower Bound, 95% conf   : 0.0
   Theta (double)          : 1.0
   Theta (long)            : 9223372036854775807
   Theta (long) hex        : 7fffffffffffffff
   EstMode?                : false
   Empty?                  : true
   Retained Entries        : 0
   Seed Hash               : 0 | 0
### END SKETCH SUMMARY


@ emptySketch.toByteArray
res2: Array[Byte] = Array(1, 3, 3, 0, 0, 30, 0, 0)

@ Sketch.wrap(Memory.wrap(Array[Byte](1, 3, 3, 0, 0, 0x1E, 0, 0)))
res3: Sketch =
### EmptyCompactSketch SUMMARY:
   Estimate                : 0.0
   Upper Bound, 95% conf   : 0.0
   Lower Bound, 95% conf   : 0.0
   Theta (double)          : 1.0
   Theta (long)            : 9223372036854775807
   Theta (long) hex        : 7fffffffffffffff
   EstMode?                : false
   Empty?                  : true
   Retained Entries        : 0
   Seed Hash               : 0 | 0
### END SKETCH SUMMARY
```

In C++ (or Python), the result is a Sketch seed hash mismatch:
```
>>> import datasketches
[...]

PR #151, opened 2020-05-28
no checking of seed hash for empty compact sketches

AlexanderSaydakov:


AlexanderSaydakov:
fixes #149

leerho:
The way we do this check in Java is more comprehensive in that an empty Theta sketch has a distinctive signature that allows us to accommodate older data that had bugs, for example the Empty or ordered flags were not set when they should have been.   Yet we had a lot of historical sketches stored this way.  Whether we need/want to burden C++ with these historical binary compatibility issues is an open question. 

You can see this code with comments [here](https://github.com/apache/incubator-datasketches-java/blob/master/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java) Lines 32-61.

AlexanderSaydakov:
Could you clarify your point please? Are you saying that some changes are needed?

PR #156, opened 2020-06-08
Add vector_of_kll_{floats,ints}_sketches to python

jmalkin:
New object in python that holds a vector of kll sketches and allows for batch updates. Confined to python at this point as it uses some numpy object types.

Maybe it should be in a separate file as it's not really a trivial wrapper around kll but something new? I'm kind of leaning towards yes.


mdhimes:
I agree with keeping it confined to Python, I'm not sure that there is much benefit (if any) to use this object with C++.

I can see benefits to keeping it within the wrapper vs. separating it into another file.  It is, ultimately, a fancy wrapper for KLL, just written in a manner that is more optimized for Python.  However, it isn't actually a wrapper for KLL, since it's a new object + a Python wrapper for that new object.  I would say that a separate file would be best from an organizational point of view.

jmalkin:
gcc is clearly unhappy with something in python...

jmalkin:
@mdhimes split out now, in the tests, too. Other than the class name, I think I the python interface was unchanged (aside from adding merge() and collapse() methods).

mdhimes:
Looks good to me, I ran some tests and it works as expected.  Great work Jon.

python/src/vector_of_kll.cpp
```diff
@@ -0,0 +1,487 @@
...
+template <typename T, typename C = std::less<T>, typename S = serde<T>>
+class vector_of_kll_sketches {
+  public:
+    static const uint32_t DEFAULT_K = kll_sketch<T, C, S>::DEFAULT_K;
+    static const uint32_t DEFAULT_D = 1;
+
+    explicit vector_of_kll_sketches(uint32_t k = DEFAULT_K, uint32_t d = DEFAULT_D);
+    vector_of_kll_sketches(const vector_of_kll_sketches& other);
+    vector_of_kll_sketches(vector_of_kll_sketches&& other) noexcept;
+    vector_of_kll_sketches<T,C,S>& operator=(const vector_of_kll_sketches& other);
```
AlexanderSaydakov, line 42: no need to spell out the template params here, but no harm either

PR #158, opened 2020-06-12
Code coverage

jmalkin:
The new coverage workflow will only run on PRs. Creating one since we'll need to merge eventually, but also in hopes of triggering it to run in the meantime.


jmalkin:
oops -- gonna fail cuz I didn't grab lcov first

jmalkin:
Results available here, not sure why no comment:
https://coveralls.io/github/apache/incubator-datasketches-cpp

PR #161, opened 2020-06-26
Fix vector_of_kll_sketches check for number of elements

mdhimes:
The current check throws an error for 2D arrays since it looks at the 0th axis for the number of elements, instead of the last axis (ndims-1).  The array indexing is coded to assume the number of elements is in the last axis, so I updated the check to match this.


jmalkin:
Looks good. Should probably have a unit test that tests 1-d, 2-d, and some failure case. I'll try to add that if you don't get to it first.

mdhimes:
Alright, got everything fixed now, and I added unit tests for 2D and 3D updates.  The 1D case is already captured in the main unit test so I left that as is.

mdhimes:
Yeah, I see now that the unchecked approach is better.  I had misinterpreted part of the pybind docs, and I thought the result of the indexing issue (seeing what looked like unallocated memory) was due to the object not being densely packed.

PR #164, opened 2020-07-21
add array input to kll sketches in python

jmalkin:


jmalkin:
Failures look like they're timeouts on the test job. I did run the tests locally on a Mac prior to committing so I'm not concerned if those fail because the queue is overloaded. And the others are fine.

PR #166, opened 2020-07-22
Issue-165: Remove c-style const casts from kll/ and common directories

gaborkaszab:
The motivation behind this change is that I recently started integrating
DataSketches KLL into Apache Impala and our Clang tidy build showed some issues
where c-style casts were used to remove const-ness of variables.


jmalkin:
Github actions seem to be having trouble with the Mac queue. Checked it locally and it was fine.

PR #170, opened 2020-08-25
Tuple sketch

AlexanderSaydakov:
This is an implementation of generic tuple sketch, specialized array of doubles sketch (with serialization compatible with Java equivalent) and experimental theta sketch on top of the same base layer (it should eventually replace the existing theta sketch). Perhaps some reorganization might help to separate this large set of files into several groups (for instance, moving the common base to a separate directory).


tuple/include/array_of_doubles_sketch_impl.hpp
```diff
@@ -0,0 +1,238 @@
...
+    const uint32_t unused32 = 0;
+    os.write(reinterpret_cast<const char*>(&unused32), sizeof(unused32));
+    for (const auto& it: this->entries_) {
+      os.write(reinterpret_cast<const char*>(&it.first), sizeof(uint64_t));
+    }
+    for (const auto& it: this->entries_) {
+      os.write(reinterpret_cast<const char*>(it.second.data()), it.second.size() * sizeof(double));
+    }
+  }
+}
```
davecromberge, line 102: Assuming sketches are compatible across implementations, do we test that the raw bytes that have been serialised are compatible with the java counterpart?

tuple/include/array_of_doubles_intersection.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <vector>
+#include <memory>
+
+#include "array_of_doubles_sketch.hpp"
+#include "tuple_intersection.hpp"
+
+namespace datasketches {
+
+template<
+  typename Policy,
```
davecromberge, line 32: For consistency purposes, is it better to use Mode here, or use Policy in the java version?  The same conceptual idea has different names in both implementations.  Do we strive for this or is this a non-issue?

tuple/include/theta_intersection_base_impl.hpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+      } else if (sketch.is_ordered()) {
+        break; // early stop
+      }
+      ++count;
+    }
+    if (count > sketch.get_num_retained()) {
+      throw std::invalid_argument(" more keys then expected, possibly corrupted input sketch");
+    } else if (!sketch.is_ordered() && count < sketch.get_num_retained()) {
+      throw std::invalid_argument(" fewer keys then expected, possibly corrupted input sketch");
```
davecromberge, line 83: There is a minor typo in the exception message `then` should be `than`

tuple/include/theta_set_difference_base_impl.hpp
```diff
@@ -0,0 +1,80 @@
...
+      for (const auto& entry: b) {
+        const uint64_t hash = EK()(entry);
+        if (hash < theta) {
+          table.insert(table.find(hash).first, hash);
+        } else if (b.is_ordered()) {
+          break; // early stop
+        }
+      }
+
+      // scan A lookup B
```
davecromberge, line 63: Is the scan optimised to choose the "smaller" of the two, assuming its cheaper to scan smaller sets?

tuple/test/tuple_sketch_allocation_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+      ++count;
+    }
+    REQUIRE(count == update_sketch.get_num_retained());
+
+    auto bytes = compact_sketch.serialize(0, test_type_serde());
+    auto deserialized_sketch = compact_tuple_sketch_test::deserialize(bytes.data(), bytes.size(), DEFAULT_SEED, test_type_serde());
+    REQUIRE(deserialized_sketch.get_estimate() == compact_sketch.get_estimate());
+
+    // update sketch copy
+//    std::cout << update_sketch.to_string();
```
davecromberge, line 82: Did you mean to this debugging commented out for future convenience?

davecromberge (commented):
I have reviewed this PR as a somebody who is not familiar with cpp, so I have not been able to verify the correctness of much of the code.  However, I think that it is neat and well structured, and I've just mainly focussed on the documentation.

In particular, I think that using a base abstraction for the theta set operations works quite well.  The sharing of the implementation between the tuple and theta sketch implementations is very effective.

tuple/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+
+set(tuple_HEADERS "")
+list(APPEND tuple_HEADERS "include/tuple_sketch.hpp;include/tuple_sketch_impl.hpp")
+list(APPEND tuple_HEADERS "include/tuple_union.hpp;include/tuple_union_impl.hpp")
+list(APPEND tuple_HEADERS "include/tuple_intersection.hpp;include/tuple_intersection_impl.hpp")
+list(APPEND tuple_HEADERS "include/tuple_a_not_b.hpp;include/tuple_a_not_b_impl.hpp")
+list(APPEND tuple_HEADERS "include/array_of_doubles_sketch.hpp;include/array_of_doubles_sketch_impl.hpp")
+list(APPEND tuple_HEADERS "include/array_of_doubles_union.hpp;include/array_of_doulbes_union_impl.hpp")
+list(APPEND tuple_HEADERS "include/array_of_doubles_intersection.hpp;include/array_of_doulbes_intersection_impl.hpp")
+list(APPEND tuple_HEADERS "include/array_of_doubles_a_not_b.hpp;include/array_of_doulbes__ot_b_impl.hpp")
```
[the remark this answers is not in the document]
AlexanderSaydakov: fixed

tuple/include/array_of_doubles_sketch_impl.hpp
AlexanderSaydakov: Yes, there are unit tests for this

tuple/include/array_of_doubles_intersection.hpp
AlexanderSaydakov: We try to be consistent if it makes sense. I think in this case the implementation is substantially different. A corresponding thing in Java ArrayOfDoublesIntersection is called Combiner. And there is no way currently to override similar combining in sketch update and in union. Here in C++ I decided to call this concept a Policy, and it can be user-defined for update, union and intersection. It defaults to sum in update and union, but there is no default for the intersection since it is not quite clear what to do with values for the same keys. The naming is debatable, of course.

tuple/include/theta_intersection_base_impl.hpp
AlexanderSaydakov: fixed

tuple/include/array_of_doubles_intersection.hpp
davecromberge: I agree with your reasons for keeping it as it is, seeing as the implementations are different. The default policy that you have chosen for the different operations also sensible.

tuple/include/theta_set_difference_base_impl.hpp
AlexanderSaydakov: I am not sure what algorithm you have in mind. We are computing a set difference here: we need to copy all entries from A except those also contained in B. So we convert B into a hash table for fast lookup, and then scan A. There is one potential optimization I see: avoid building a hash table from B if it happens to be a hash table already (an update sketch). I did not want to introduce a special treatment and somehow penetrate into B to figure out what it is for a limited (in my view) use case. I believe that the most typical scenario is that sketches are converted to compact form right after building them.

tuple/test/tuple_sketch_allocation_test.cpp
AlexanderSaydakov: We can clean this up, but I don't think it is important in tests. I see your point, and I prefer not to keep commented-out lines in the main code

tuple/include/theta_set_difference_base_impl.hpp
davecromberge: You are correct, a different algorithm came to mind when I saw the comment on line 63 without me sufficiently appreciating the context.  I recall the java algorithm for intersections having an optimisation about scan order in the data tables, which is not applicable here.
I agree with you about avoiding the special treatment for the update sketch.

tuple/include/array_of_doubles_sketch.hpp
```diff
@@ -0,0 +1,179 @@
...
+class compact_array_of_doubles_sketch_alloc: public compact_tuple_sketch<aod<A>, AllocAOD<A>> {
+public:
+  using Base = compact_tuple_sketch<aod<A>, AllocAOD<A>>;
+  using Entry = typename Base::Entry;
+  using AllocEntry = typename Base::AllocEntry;
+  using AllocU64 = typename Base::AllocU64;
+  using vector_bytes = typename Base::vector_bytes;
+
+  static const uint8_t SERIAL_VERSION = 1;
+  static const uint8_t SKETCH_FAMILY = 9;
```
jmalkin, line 149: Not thrilled about these being redefined here when they're already in tuple_sketch.  Might be even better if we could have some of these defined in a common file shared across sketches. But not a blocker.

jmalkin (approved):
I'd do a few minor things differently stylistically, and I admit that my mind couldn't handle reading all the tests. I tried to follow through the logic on the base hash table, and everything else was in effect a wrapper around that common table.  Overall I really like the design concept, even if there's a small performance hit for the experimental theta vs the original version.

tuple/include/theta_update_sketch_base_impl.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+// builder
+
+template<typename Derived, typename Allocator>
+theta_base_builder<Derived, Allocator>::theta_base_builder(const Allocator& allocator):
+allocator_(allocator), lg_k_(DEFAULT_LG_K), rf_(DEFAULT_RESIZE_FACTOR), p_(1), seed_(DEFAULT_SEED) {}
+
+template<typename Derived, typename Allocator>
+Derived& theta_base_builder<Derived, Allocator>::set_lg_k(uint8_t lg_k) {
+  if (lg_k < MIN_LG_K) {
```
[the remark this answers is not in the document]
AlexanderSaydakov: I think so. I will add that

tuple/include/theta_update_sketch_base_impl.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+template<typename Derived, typename Allocator>
+Derived& theta_base_builder<Derived, Allocator>::set_resize_factor(resize_factor rf) {
+  rf_ = rf;
+  return static_cast<Derived&>(*this);
+}
+
+template<typename Derived, typename Allocator>
+Derived& theta_base_builder<Derived, Allocator>::set_p(float p) {
+  if (p < 0 || p > 1) throw std::invalid_argument("sampling probability must be between 0 and 1");
```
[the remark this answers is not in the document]
AlexanderSaydakov: I am not sure it makes much sense. Sampling with probability of 0?

tuple/include/array_of_doubles_sketch.hpp
AlexanderSaydakov: This is a different family and different serial format. I don't really like this concept, but it serves as a compatibility check.

jmalkin:
The last few edits look good. Merge away!

PR #176, opened 2020-12-22
Req sketch

AlexanderSaydakov:


leerho:
Let's wait on merging this to master until we hear from Pavel.

leerho:
I am reviewing with Pavel a small change in the "fixRseFactor" from .06 to .084.  If he feels comfortable with this I will make this change in Java and you can make the same change in C++. Then we will be ready to release.

jmalkin:
I did nothing productive over the weekend so I still need to finish the review of this.

req/include/req_sketch.hpp
```diff
@@ -0,0 +1,395 @@
...
+  const_iterator& operator++();
+  const_iterator& operator++(int);
+  bool operator==(const const_iterator& other) const;
+  bool operator!=(const const_iterator& other) const;
+  std::pair<const T&, const uint64_t> operator*() const;
+private:
+  using CompactorsIterator = typename std::vector<Compactor, AllocCompactor>::const_iterator;
+  CompactorsIterator compactors_it_;
+  CompactorsIterator compactors_end_;
+  const T* compactor_it_;
```
jmalkin, line 386: compactors_it_ vs compactor_it_ seems like it's asking for trouble. I'd probably change the first to something like levels_it_ to avoid future confusion.

jmalkin (approved):
Way too much code so I'm not going to pretend I would have caught anything particuarly subtle, but seems ok

req/include/req_sketch.hpp
AlexanderSaydakov: renamed to levels

PR #178, opened 2021-01-08
Remove incubat*

leerho:
Alex, I need you to do the same in Default/ and local/ as required. - thanks.


jmalkin (approved):
I'm pretty sure Default/ and local/ are from your IDE.

PR #181, opened 2021-01-14
support passing an allocator instance

AlexanderSaydakov:


jmalkin (approved):
Allocator as well as typedef -> using and using trailing _ for member variables. I didn't separately search the new branch to check for any remaining instances of non-instantiated allocator usage. Looks fine otherwise.

PR #183, opened 2021-01-19
support allocator instance

AlexanderSaydakov:


cpc/include/cpc_sketch_impl.hpp
```diff
@@ -401,7 +406,7 @@ string<A> cpc_sketch_alloc<A>::to_string() const {
 
 template<typename A>
 void cpc_sketch_alloc<A>::serialize(std::ostream& os) const {
-  compressed_state<A> compressed;
+  compressed_state<A> compressed(A(sliding_window.get_allocator()));
```
jmalkin, line 409: Should serialize() have the option of using a different allocator? It's not entirely obvious that, say, a limited-size allocator would want transient things like serialization to complete for space with the main stored data?

jmalkin (approved):
Serialize using the same allocator certainly isn't a blocker, just something to consider. LGTM otherwise.

cpc/include/cpc_sketch_impl.hpp
AlexanderSaydakov: We are using a provided allocator for all allocations within an instance, including temporary allocations. I think this is good enough.

PR #185, opened 2021-01-27
Matches the Theta interface of the Java implementation

chufucun:
I recently started integrating DataSketches Theta into Apache Impala and need to add matching methods in the Java implementation.
https://github.com/apache/datasketches-java/blob/master/src/main/java/org/apache/datasketches/theta/Union.java

### DIFF
--- a/theta/include/theta_union.hpp
+++ b/theta/include/theta_union.hpp
@@ -48,6 +48,95 @@ public:
   void update(const theta_sketch_alloc<A>& sketch);
 
   /**
+   * This method is to update the union with a given string.
+   * @param value string to update the union with
+   */
+  void update(const std::string &value);
+  
+  /**
+   * This method is to update the union with a given unsigned 64-bit integer.
+   * @param value uint64_t to update the union with
+   */
+  void update(uint64_t value);
+  
+  /**
+   * This method is to update the union with a given signed 64-bit integer.
+   * @param value int64_t to update the union with
+   */
+  void update(int64_t value);
+  
+  /**
+   * This method is to update the union with a given unsigned 32-bit integer.
+   * For compatibility with Java implementation.
+   * @param value uint32_t to update the union with
+   */
+  void update(uint32_t value);
+  
+  /**
+   * This method is to update the union with a given signed 32-bit integer.
+   * For compatibility with Java implementation.
+   * @param value int32_t to update the union with
+   */
+  void update(int32_t value);
+  
+  /**
+   * This method is to update the union with a given unsigned 16-bit integer.
+   * For compatibility with Java implementation.
+   * @param value uint16_t to update the union with
+   */
+  void update(uint16_t value);
+  
+  /**
+   * This method is to update the union with a given signed 16-bit integer.
+   * For compatibility with Java implementation.
+   * @param value int16_t to update the union with
+   */
+  void update(int16_t value);
+  
+  /**
+   * This method is to update the union with a given unsigned 8-bit integer.
+   * For compatibility with Java implementation.
+   * @param value uint8_t to update the union with
+   */
+  void update(uint8_t value);
+  
+  /**
+   * This method is to update the union with a given signed 8-bit integer.
+   * For compatibility with Java implementation.
+   * @param value int8_t to update the union with
+   */
+  void update(int8_t value);
+  
+  /**
+   * This method is to update the union with a given double-precision floating point value.
+   * For compatibility with Java implementation.
+   * @param value double to update the union with
+   */
+  void update(double value);
+  
+  /**
+   * This method is to update the union with a given floating point value.
+   * For compatibility with Java implementation.
+   * @param value float to update the union with
+   */
+  void update(float value);
+  
+  /**
+   * This method is to update the union with given data of any type.
+   * This is a "universal" update that covers all cases above,
+   * but may produce different hashes.
+   * Be very careful to hash input values consistently using the same approach
+   * both over time and on different platforms
+   * and while passing sketches between C++ environment and Java environment.
+   * Otherwise two sketches that should represent overlapping sets will be disjoint
+   * For instance, for signed 32-bit values call update(int32_t) method above,
+   * which does widening conversion to int64_t, if compatibility with Java is expected
+   * @param data pointer to the data
+   * @param length of the data in bytes
+   */
+  void update(const void *data, unsigned length);
+
+  /**
    * This method produces a copy of the current state of the union as a compact sketch.
    * @param ordered optional flag to specify if ordered sketch should be produced
    * @return the result of the union
--- a/theta/include/theta_union_impl.hpp
+++ b/theta/include/theta_union_impl.hpp
@@ -50,6 +50,79 @@ void theta_union_alloc<A>::update(const theta_sketch_alloc<A>& sketch) {
 }
 
 template<typename A>
+void theta_union_alloc<A>::update(const std::string &value) {
+    if (value.empty()) return;
+    is_empty_ = false;
+    state_.update(value);
+}
+
+template<typename A>
+void theta_union_alloc<A>::update(uint64_t value){
+    is_empty_ = false;
+    state_.update(value);
+}
+  
+template<typename A>
+void theta_union_alloc<A>::update(int64_t value){
+    is_empty_ = false;
+    state_.update(value);
+}
+  
+template<typename A>
+void theta_union_alloc<A>::update(uint32_t value){
+    is_empty_ = false;
+    state_.update(value);
+}
+  
+template<typename A>
+void theta_union_alloc<A>::update(int32_t value){
+    is_empty_ = false;
+    state_.update(value);
+}
+  
+template<typename A>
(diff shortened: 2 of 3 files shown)


AlexanderSaydakov:
It seems that this change breaks something in Python wrapper.
I never liked this mixed union functionality. In my view it is just unnecessary complexity. That is why I did not implement it in C++.
There is a complication: we have a new implementation of Theta sketch, which was developed as a part of Tuple sketch effort. The goal was to have a common base for both Theta and Tuple sketches. Currently this new Theta sketch implementation is sitting under Tuple sketch as experimental. I think it is time to make the move: replace the existing Theta sketch with this new implementation. I think it would be better to make sure that your integration works with this new version.
Perhaps we could add these update methods to the Theta and Tuple unions, but I still think that they are not necessary.

chufucun:
> It seems that this change breaks something in Python wrapper.
> I never liked this mixed union functionality. In my view it is just unnecessary complexity. That is why I did not implement it in C++.
> There is a complication: we have a new implementation of Theta sketch, which was developed as a part of Tuple sketch effort. The goal was to have a common base for both Theta and Tuple sketches. Currently this new Theta sketch implementation is sitting under Tuple sketch as experimental. I think it is time to make the move: replace the existing Theta sketch with this new implementation. I think it would be better to make sure that your integration works with this new version.
> Perhaps we could add these update methods to the Theta and Tuple unions, but I still think that they are not necessary.

thx reply.
Use udaf function to integrate theta into impala. This function mainly affects the function logic of the merge phase.
I found a solution, but the code doesn't look very elegant, I will test the performance and reply again。

**Merge function:**
The `dst` parameter is used to store intermediate results and is initialized in the init phase

1. use update_theta_sketch and theta_union
```c++
void AggregateFunctions::DsThetaMerge(
    FunctionContext* ctx, const StringVal& src, StringVal* dst) {
  DCHECK(!src.is_null);
  DCHECK(!dst->is_null);
  DCHECK(dst->len == sizeof(datasketches::update_theta_sketch)
      or dst->len == sizeof(datasketches::compact_theta_sketch));
  auto src_sketch = datasketches::theta_sketch::deserialize((void*)src.ptr, src.len);
  if (src_sketch->is_empty()) return;

  auto dst_sketch_ptr = reinterpret_cast<datasketches::theta_sketch*>(dst->ptr);

  datasketches::theta_union union_sketch = datasketches::theta_union::builder().build();
  union_sketch.update(*src_sketch);
  union_sketch.update(*dst_sketch_ptr);

  // Note, Union_result may be inconsistent with dst and needs to reallocate space
  const datasketches::compact_theta_sketch union_result = union_sketch.get_result();
  ctx->Reallocate(dst->ptr, sizeof(datasketches::compact_theta_sketch));
  if (LIKELY(dst->ptr != NULL)) {
    memset(dst->ptr, 0, sizeof(datasketches::compact_theta_sketch));
    datasketches::compact_theta_sketch* sketch_ptr =
        reinterpret_cast<datasketches::compact_theta_sketch*>(dst->ptr);
    std::uninitialized_fill_n(sketch_ptr, 1, union_result);
  } else {
    DCHECK(!ctx->impl()->state()->GetQueryStatus().ok());
  }
}
```
2. use theta_union(need add new update functionality)
```c++
void AggregateFunctions::DsThetaMerge(
    FunctionContext* ctx, const StringVal& src, StringVal* dst) {
  DCHECK(!src.is_null);
  DCHECK(!dst->is_null);
  DCHECK_EQ(dst->len, sizeof(datasketches::theta_union));

  // Note, 'src' is a serialized compact_theta_sketch and not a serialized theta_union.
  auto src_sketch =
      datasketches::compact_theta_sketch::deserialize((void*)src.ptr, src.len);

  datasketches::theta_union* dst_union_ptr =
      reinterpret_cast<datasketches::theta_union*>(dst->ptr);

  dst_union_ptr->update(src_sketch);
}
```

AlexanderSaydakov:
I don't understand your point. Is this code supposed to justify the need of the update methods added in this pull request? I don't see them being called.

chufucun:
The reason for submitting this pull is that the intermediate result (dst) can only initialize one type (eg: update_theta_sketch), and theta_union.get_result() returns compact_theta_sketch. This type is inconsistent with the initialization type (update_theta_sketch), and the merge phase needs to reallocate space each time. 

Observing that [datasketches-hive](https://github.com/apache/datasketches-hive) is implemented with union, which led to the idea of adding a new update function to theta_union and using theta_union to implement the `ds_theta_sketch()`.
https://github.com/apache/datasketches-hive/blob/master/src/main/java/org/apache/datasketches/hive/theta/DataToSketchUDAF.java#L126
https://github.com/apache/datasketches-hive/blob/master/src/main/java/org/apache/datasketches/hive/theta/UnionState.java#L66

The new update function added in theta_union is used in the **update phase**。

Complete Code:
Note: use `update_theta_sketch` is similar to this but the merge phase code is different.

```c++
// Init phase：Initialize intermediate results (dst) for merge and update phase
void AggregateFunctions::DsThetaInit(FunctionContext* ctx, StringVal* dst) {
  // Allocate space
  AllocBuffer(ctx, dst, sizeof(datasketches::theta_union));
  if (UNLIKELY(dst->is_null)) {
    DCHECK(!ctx->impl()->state()->GetQueryStatus().ok());
    return;
  }
  // Constructs object initialization memor
  datasketches::theta_union* sketch_ptr =  reinterpret_cast<datasketches::theta_union*>(dst->ptr);
  datasketches::theta_union sketch = datasketches::theta_union::builder().build();
  std::uninitialized_fill_n(sketch_ptr, 1, sketch);
}

// Update phase
template <typename T>
void AggregateFunctions::DsThetaUpdate(FunctionContext* ctx, const T& src, StringVal* dst) {
  if (src.is_null) return;
  DCHECK(!dst->is_null);
  DCHECK_EQ(dst->len, sizeof(datasketches::theta_union));
  datasketches::theta_union* sketch_ptr = reinterpret_cast<datasketches::theta_union*>(dst->ptr);
  // need add new update functionality in theta_union   
  sketch_ptr->update(src.val);
}

// Serialize phase
StringVal AggregateFunctions::DsThetaSerialize(FunctionContext* ctx, const StringVal& src) {
  DCHECK(!src.is_null);
  DCHECK_EQ(src.len, sizeof(datasketches::theta_union));
  datasketches::theta_union* union_ptr = reinterpret_cast<datasketches::theta_union*>(src.ptr);
  StringVal dst = SerializeDsThetaUnion(ctx, *union_ptr);
  ctx->Free(src.ptr);
  return dst;
}

// Merge phase
void AggregateFunctions::DsThetaMerge(FunctionContext* ctx, const StringVal& src, StringVal* dst) {
  DCHECK(!src.is_null);
  DCHECK(!dst->is_null);
  DCHECK_EQ(dst->len, sizeof(datasketches::theta_union));

  // Note, 'src' is a serialized compact_theta_sketch and not a serialized theta_union.
  auto src_sketch =  datasketches::compact_theta_sketch::deserialize((void*)src.ptr, src.len);

  datasketches::theta_union* dst_union_ptr = reinterpret_cast<datasketches::theta_union*>(dst->ptr);

  dst_union_ptr->update(src_sketch);
}

// Finalize phase, return estimate value.
BigIntVal AggregateFunctions::DsThetaFinalize(FunctionContext* ctx, const StringVal& src) {
  DCHECK(!src.is_null);
  DCHECK_EQ(src.len, sizeof(datasketches::theta_union));
  datasketches::theta_union* sketch_ptr = reinterpret_cast<datasketches::theta_union*>(src.ptr);
  auto sketch = sketch_ptr->get_result();
  ctx->Free(src.ptr);
  return sketch.get_estimate();
}
```

AlexanderSaydakov:
Why do you serialize union? We are removing this functionality from the new version as well as serialization of update sketch. Only compact sketches can be serialized. I am afraid I don't quite get your operation sequence. I would think that in the update phase the state of the aggregation should be update_theta_sketch, which should be converted to a compact_theta_sketch at the end of the phase, serialized and shipped to the merge phase. And the merge phase should only receive compact sketches to union them.

AlexanderSaydakov:
We have just promoted the experimental theta sketch implementation to replace the previous one. This creates a conflict with your proposed changes. Sorry for this confusion. I would suggest closing this request for now. If you really need this mixed union functionality, we can discuss it. However I still believe that we can easily do without. I still don't understand why you gave some examples above that don't show calling these methods.
New implementation allows serializing compact sketches only. I strongly believe that serializing updatable sketches or unions should not be necessary. If you are facing some difficulty in Impala, let us discuss this. I suggest opening an issue for that. We want to help you. Thank you.

chufucun:
Sorry about the delay in responding here. I've been busy with other things.

The serialized union is actually the result of `theta_union.get_result()`.
```c++
StringVal SerializeDsThetaUnion(
    FunctionContext* ctx, const datasketches::theta_union& ds_union) {
  std::stringstream serialized_input;
  datasketches::compact_theta_sketch sketch = ds_union.get_result();
  return SerializeDsThetaSketch(ctx, sketch);
}
```
The problem was resolved using the existing interface and the performance met the requirements.

PR #189, opened 2021-01-29
promoted new implementation of theta sketch

AlexanderSaydakov:
This is a new implementation of theta sketch which shared the base layer with tuple sketch. This version of theta sketch was developed a while ago together with tuple sketch. This pull request is to replace the existing theta sketch with the new version. Serialization was changed: in the new version only compact sketches can be serialized and deserialized.


jmalkin (approved):
I'd look through tuple before. I re-read code for the _base versions and then looked at the theta wrappers around them, trying to make sure i understood differences between theta and tuple.

There's a lot of code and it's pretty complexly nested, so the approval should be taken with the usual caveats for any changes this large.

jmalkin:
Also, I dunno why the CI build is broken, but it works locally and I'm confident the breakage isn't specific to this change.

PR #195, opened 2021-02-12
test that allocator default constructor is not called

AlexanderSaydakov:


hll/test/HllSketchTest.cpp
```diff
@@ -162,27 +162,23 @@ TEST_CASE("hll sketch: check ser sizes", "[hll_sketch]") {
...
-    for (int i = 25; i < (1 << 20); ++i) { sk.update(i); }
-    oss << sk.to_string(false, true, true, true);
-    oss << sk.to_string(false, true, true, false);
-
-    sk = hll_sketch_test_alloc(8, HLL_8);
-    for (int i = 0; i < 25; ++i) { sk.update(i); }
-    oss << sk.to_string(false, true, true, true);
-  }
-  REQUIRE(test_allocator_total_bytes == 0);
+  hll_sketch sk(15, HLL_4);
```
jmalkin, line 165: Seems reasonable to have some non-test_allocator usage, too, but why was it explicitly removed in this case?

sampling/test/var_opt_allocation_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+  REQUIRE(u.get_result().get_n() == 2 * n);
+
+  // move constructor
+  var_opt_union<test_type> u2(std::move(u));
+  REQUIRE(u2.get_result().get_n() == 2 * n);
+
+  // move assignment
+  var_opt_union<test_type> u3(k);
+  u3 = std::move(u2);
+  REQUIRE(u3.get_result().get_n() == 2 * n);
```
jmalkin, line 86: Do we want to check for total bytes allocated here?

jmalkin (commented):
Applying the idea that unit tests can also help people understand code usage, do we want to have some examples of basic usage that don't use the test allocator? That ends up requiring the use of all arguments in the constructors, sometimes including values people wouldn't usually need to care about.

jmalkin:
I guess the idea of tutorial tests is probably outside the scope of this specific PR.

hll/test/HllSketchTest.cpp
AlexanderSaydakov: Yes, I thought about splitting this set of tests into simple tests that can serve as usage examples, and more elaborate allocation tests with this additional clutter (passing allocators, checking allocated bytes and so on). I am afraid it is a bit more involved than I planned at this moment.

sampling/test/var_opt_allocation_test.cpp
AlexanderSaydakov: Right, this particular test was moved as-is from var_opt_union_test.cpp because the test_type (as it currently is) can be included only in one file per module.

hll/test/HllSketchTest.cpp
jmalkin: Fair. Gigantic PRs are obnoxious in many ways and we don't need them all to do everything. Thanks.

sampling/test/var_opt_allocation_test.cpp
jmalkin: Got it. Thanks.

jmalkin (approved):
Any questions addressed. Looks good.

jmalkin (approved):
👍

PR #201, opened 2021-02-26
added missing guards

AlexanderSaydakov:


jmalkin (approved):
Should the new test I added have caught this? Anyway, good to notice it.

PR #202, opened 2021-02-26
Readme update compiler versions

AlexanderSaydakov:


jmalkin:
Our CI job used:
  * Mac: `AppleClang 12.0.0.12000032`
  * Ubuntu: `GNU 7.5.0`
  * Windows: `MSVC 19.28.29337.0`

I don't think we've tested on 4.8.5 recently?

AlexanderSaydakov:
Do you think it makes much sense to maintain these versions?

jmalkin:
With the CI job, maybe not. We should probably do a 4.8.5 check occasionally but beyond that I think we can probably just kill this.

AlexanderSaydakov:
I agree, let's just remove this part.

jmalkin (approved):

PR #204, opened 2021-03-10
Unify serialization

AlexanderSaydakov:
Simplified and unified serialization and deserialization except HLL, where it is quite different form everything else, and will take a separate effort.

### DIFF
--- a/kll/include/kll_sketch_impl.hpp
+++ b/kll/include/kll_sketch_impl.hpp
@@ -388,28 +388,28 @@ template<typename T, typename C, typename S, typename A>
 void kll_sketch<T, C, S, A>::serialize(std::ostream& os) const {
   const bool is_single_item = n_ == 1;
   const uint8_t preamble_ints(is_empty() || is_single_item ? PREAMBLE_INTS_SHORT : PREAMBLE_INTS_FULL);
-  os.write(reinterpret_cast<const char*>(&preamble_ints), sizeof(preamble_ints));
+  write(os, preamble_ints);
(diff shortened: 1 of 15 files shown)


jmalkin:
I think we need to figure out what happened to code coverage.

kll/include/kll_sketch_impl.hpp
```diff
@@ -388,28 +388,28 @@ template<typename T, typename C, typename S, typename A>
...
-    os.write((char*)&n_, sizeof(n_));
-    os.write((char*)&min_k_, sizeof(min_k_));
-    os.write((char*)&num_levels_, sizeof(num_levels_));
-    os.write((char*)&unused, sizeof(unused));
-    os.write((char*)levels_.data(), sizeof(levels_[0]) * num_levels_);
+    write(os, n_);
+    write(os, min_k_);
+    write(os, num_levels_);
+    write(os, unused);
+    os.write(reinterpret_cast<const char*>(levels_.data()), sizeof(levels_[0]) * num_levels_);
```
jmalkin, line 412: Feels weird to have an os.write in the middle of this

AlexanderSaydakov: this is a different case. maybe we can have another template for this as well

jmalkin (approved):
I like how it collapses the
```
declare variable
is.read(reinterpret_cast<>())
```
code blocks into something really easy to read.

I think it's eventually worth adding a template for the remaining cases, but this is still a nice improvement.

PR #208, opened 2021-03-22
fixed iterator

### LINKED ISSUE #205: Empty kll sketches behavior
Hi! I've been experimenting with the kll sketch lately, and I don't understand the behavior of the ["empty"](https://github.com/apache/datasketches-cpp/blob/master/kll/test/kll_sketch_test.cpp#L57) test.

On [line 73](https://github.com/apache/datasketches-cpp/blob/master/kll/test/kll_sketch_test.cpp#L73) I see a counting loop and a `REQUIRE(count == 0);` that in my mind means "it should never enter the loop", but I was surprised that on my machine (Windows 10, Visual Studio 19 Community latest version, both Release and Debug configuration, x64 platform) that is not the case: it just loops _until count overflows_ and than exits the loop. This makes the test "pass", but (apart from overflow being UB) I don't think it's the right behavior.

To make sure I'm not missing something, I changed the test with
```diff
diff --git a/kll/test/kll_sketch_test.cpp b/kll/test/kll_sketch_test.cpp
index def96f6..c29c135 100644
--- a/kll/test/kll_sketch_test.cpp
+++ b/kll/test/kll_sketch_test.cpp
@@ -74,6 +74,7 @@ TEST_CASE("kll sketch", "[kll_sketch]") {
     for (auto& it: sketch) {
       (void) it; // to suppress "unused" warning
       ++count;
+      FAIL("the sketch should be empty");
     }
     REQUIRE(count == 0);
 }
```

And the test immediately fails.

I also tested this behavior using GCC (on a WSL2 machine, Ubuntu 20.04, g++  9.3.0) and the test fails, too.

Am I missing something?

AlexanderSaydakov:
fixes #205

### DIFF
--- a/kll/include/kll_sketch_impl.hpp
+++ b/kll/include/kll_sketch_impl.hpp
@@ -1059,14 +1059,14 @@ typename kll_sketch<T, C, S, A>::const_iterator kll_sketch<T, C, S, A>::begin() 
 
 template <typename T, typename C, typename S, typename A>
 typename kll_sketch<T, C, S, A>::const_iterator kll_sketch<T, C, S, A>::end() const {
-  return kll_sketch<T, C, S, A>::const_iterator(nullptr, nullptr, num_levels_);
+  return kll_sketch<T, C, S, A>::const_iterator(nullptr, levels_.data(), num_levels_);
 }
 
 // kll_sketch::const_iterator implementation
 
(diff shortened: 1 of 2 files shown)


jmalkin (approved):
My understanding is that If index is null then you have an end pointer, and you set the values appropriately, otherwise you start at the beginning.

If that's correct, looks good.

PR #217, opened 2021-05-19
added seed

### LINKED ISSUE #216: Jaccard similarity doesn't handle custom seeds
If two sketches with custom seeds are loaded and passed to `theta_jaccard_similarity::jaccard()`, an exception is generated for `seed hash mismatch`. It looks like the problem is that Jaccard builds unions and intersections but doesn't have a way to pass the relevant see to those sketches.

AlexanderSaydakov:
fixes #216

### DIFF
--- a/theta/test/theta_jaccard_similarity_test.cpp
+++ b/theta/test/theta_jaccard_similarity_test.cpp
@@ -100,6 +100,28 @@ TEST_CASE("theta jaccard: half overlap estimation mode", "[theta_sketch]") {
   REQUIRE(jc[2] == Approx(0.33).margin(0.01));
 }
 
+TEST_CASE("theta jaccard: half overlap estimation mode custom seed", "[theta_sketch]") {
+  const uint64_t seed = 123;
+  auto sk_a = update_theta_sketch::builder().set_seed(seed).build();
+  auto sk_b = update_theta_sketch::builder().set_seed(seed).build();
+  for (int i = 0; i < 10000; ++i) {
+    sk_a.update(i);
+    sk_b.update(i + 5000);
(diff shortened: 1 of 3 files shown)


theta/test/theta_jaccard_similarity_test.cpp
```diff
@@ -140,4 +179,21 @@ TEST_CASE("theta jaccard: dissimilarity test", "[theta_sketch]") {
...
+  const uint64_t seed = 1234;
+
+  auto expected = update_theta_sketch::builder().set_lg_k(min_lg_k).set_seed(seed).build();
+  for (int i = 0; i < u1; ++i) expected.update(i);
+
+  auto actual = update_theta_sketch::builder().set_lg_k(min_lg_k).set_seed(seed).build();
+  for (int i = 0; i < u2; ++i) actual.update(i);
+
+  REQUIRE(theta_jaccard_similarity::dissimilarity_test(actual, expected, threshold, seed));
+  REQUIRE_FALSE(theta_jaccard_similarity::dissimilarity_test(actual, actual, threshold));
```
jmalkin, line 196: should this line have the seed passed in, too?

jmalkin: oh, nevermind, misread

jmalkin:
👍

jmalkin (approved):

theta/test/theta_jaccard_similarity_test.cpp
AlexanderSaydakov: yes. the test did not fail because the implementation checks for identical input upfront

PR #218, opened 2021-05-26
Theta fix and cleanup

AlexanderSaydakov:
- fixed union base (I think this bug could lead to slightly incorrect results in rare circumstances)
- minor cleanup and slight performance improvement


jmalkin (approved):
Not obvious to me when the table theta and union theta will disagree, but it at least seems quite reasonable to check both before inserting.

and +1 on finding a way to avoid that `insert()` call

PR #220, opened 2021-06-07
changed policy to return void

AlexanderSaydakov:
Some early development version of the intersection base expected a value from an intersection policy. At some point during development this was changed to void, but this policy in the Theta sketch still returns a value in a void context. It does not affect the result, but potentially is more costly and a bit confusing.


will-lauer (approved):

jmalkin (approved):
The only place where policy_ is obviously used in the impl does not use a result, so this seems correct.

PR #222, opened 2021-06-16
CPC max serialized size bytes

AlexanderSaydakov:


will-lauer (approved):

will-lauer (approved):

cpc/include/cpc_sketch_impl.hpp
```diff
@@ -681,6 +679,46 @@ cpc_sketch_alloc<A> cpc_sketch_alloc<A>::deserialize(const void* bytes, size_t s
...
+    39532,  // lg_k = 16
+    78880,  // lg_k = 17
+    157516, // lg_k = 18
+    314656  // lg_k = 19
+};
+
+template<typename A>
+size_t cpc_sketch_alloc<A>::get_max_serialized_size_bytes(uint8_t lg_k) {
+  check_lg_k(lg_k);
+  if (lg_k <= 19) return CPC_EMPIRICAL_MAX_SIZE_BYTES[lg_k - 4] + 40;
```
jmalkin, line 710: Might be clearer to use CPC_MIN_LG_K instead of 4. And should we define a preamble size constant?

jmalkin (approved):

PR #223, opened 2021-06-23
wrapped compact theta sketch

AlexanderSaydakov:


AlexanderSaydakov:
Theta union lgk16 wrapped compact

This is a comparison of union update time with compact sketches ready at hand, compact sketches with deserialize time included, and wrapped compact sketches (effectively avoiding some cost of deserialization).
We can see that performing a union of wrapped sketches is even slightly faster than using compact sketch objects ready at hand presumably because of a faster iterator over the wrapped sketches.

theta/include/compact_theta_sketch_parser_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+    theta = reinterpret_cast<const uint64_t*>(ptr)[COMPACT_SKETCH_THETA_U64];
+  }
+  if (reinterpret_cast<const uint8_t*>(ptr)[COMPACT_SKETCH_PRE_LONGS_BYTE] == 1) {
+    return {false, true, seed_hash, 1, theta, reinterpret_cast<const uint64_t*>(ptr) + COMPACT_SKETCH_SINGLE_ENTRY_U64};
+  }
+  const uint32_t num_entries = reinterpret_cast<const uint32_t*>(ptr)[COMPACT_SKETCH_NUM_ENTRIES_U32];
+  const size_t entries_start_u64 = has_theta ? COMPACT_SKETCH_ENTRIES_ESTIMATION_U64 : COMPACT_SKETCH_ENTRIES_EXACT_U64;
+  const uint64_t* entries = reinterpret_cast<const uint64_t*>(ptr) + entries_start_u64;
+  const size_t expected_size_bytes = (entries_start_u64 + num_entries) * sizeof(uint64_t);
+  if (size < expected_size_bytes)
```
jmalkin, line 52: would we want to count too large a size as an error?

theta/include/compact_theta_sketch_parser_impl.hpp
```diff
@@ -0,0 +1,69 @@
...
+    throw std::invalid_argument(std::to_string(expected_size_bytes)
+        + " bytes expected, actual " + std::to_string(size) + ", sketch dump: " + hex_dump(reinterpret_cast<const uint8_t*>(ptr), size));
```
jmalkin, line 54: this dump can be quite large (MBs) in some cases.  not sure if we need to consider the data security risk in that it'll happily dump arbitrary amounts of memory given a pointer and size

theta/test/theta_a_not_b_test.cpp
```diff
@@ -167,6 +167,70 @@ TEST_CASE("theta a-not-b: estimation mode half overlap", "[theta_a_not_b]") {
...
+private:
+  const char* ptr;
+  size_t sz;
+};
+
+const std::vector<char> clone(std::vector<char>& other) {
+  return std::vector<char>(other);
+}
+
+TEST_CASE() {
```
jmalkin, line 224: do we need this test case and its supporting functions?

jmalkin (commented):
maybe not a big deal, but none of the tests override the default seed to explicitly test for seed mismatch

theta/include/compact_theta_sketch_parser_impl.hpp
AlexanderSaydakov: I am not sure. I can imagine a larger record and a user who does not know the size of the sketch. The primary goal here was to avoid reading past the end.

theta/test/theta_a_not_b_test.cpp
AlexanderSaydakov: Sorry, this was left there accidentally after some experimenting.

jmalkin (approved):

jmalkin:
👍

PR #230, opened 2021-08-09
Added count sketch

c-dickens:
Added a python implementation of count sketch for frequency estimation and heavy hitter detection.

### DIFF
--- a/python/src/count_sketch.py
+++ b/python/src/count_sketch.py
@@ -1,0 +1,240 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


python/src/count_sketch.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Returns the total weight inserted on the stream.
+        Is equivalent to the mass inserted over the stream, or the ell_1 norm of the
+        underlying frequency vector.
+        """
+        return self.total_weight
+
+    def is_empty(self):
+        """
+        Returns True if the sketch self.table is empty and is False otherwise.
```
jmalkin, line 207: There's a valid philosophical debate over whether adding and then subtracting an item means a sketch is truly empty. It might seem trivial with a single item, but imagine merging 2 sketches that were created with the same items but opposite weights -- if we treat it as empty, we lose the information that it has ingested data.

Certainly open to the idea that we don't care about a distinction, but it's worth thinking about what we want to capture by "empty"

jmalkin (approved):
This is a reasonable start, but I think there's still a fair amount to do before it's production ready.
- there's no serialization, so it's not yet suitable for distributed use
- merge doesn't check for sketch compatibility, including matching table sizes.  in order to be meaningful, I think we'd also want the hash functions to match?
- we really need to figure out a better hash function, as that seems important for ingesting arbitrary(-ish, in python) input types.

But I think this is sufficient to pull into a branch so we can start collaborative improvements

PR #250, opened 2021-11-02
Theta and Tuple reset

AlexanderSaydakov:


tuple/test/tuple_union_test.cpp
```diff
@@ -81,6 +81,13 @@ TEST_CASE("tuple_union float: simple case", "[tuple union]") {
   u.update(update_sketch2);
   auto result = u.get_result();
   REQUIRE(result.get_num_retained() == 3);
+
+  u.reset();
+  result = u.get_result();
+  REQUIRE(result.is_empty());
+  REQUIRE(result.get_num_retained() == 0);
+  REQUIRE(!result.is_estimation_mode());
```
jmalkin, line 89: this is `REQUIRE_FALSE` in the other tests, but not worth changing

jmalkin (approved):

PR #252, opened 2021-12-01
Theta and Tuple sketch setops sync with Java

AlexanderSaydakov:
ported tests from Java, adjusted logic


theta/include/theta_set_difference_base_impl.hpp
```diff
@@ -36,7 +36,7 @@ seed_hash_(compute_seed_hash(seed))
 template<typename EN, typename EK, typename CS, typename A>
 template<typename FwdSketch, typename Sketch>
 CS theta_set_difference_base<EN, EK, CS, A>::compute(FwdSketch&& a, const Sketch& b, bool ordered) const {
-  if (a.is_empty() || a.get_num_retained() == 0 || b.is_empty()) return CS(a, ordered);
+  if (a.is_empty() || (a.get_num_retained() > 0 && b.is_empty())) return CS(a, true);
```
jmalkin, line 39: CS(a, true) will return a sorted compact sketch regardless of the value of the `ordered` input parameter. Not really wrong with a CompactSketch, but depending on usage it may force an unnecessary sort.

leerho (approved):
Looks good to me.

PR #253, opened 2021-12-07
Adding support for older serialization versions of compact theta

will-lauer:


theta/test/theta_sketch_test.cpp
```diff
@@ -214,6 +244,78 @@ TEST_CASE("theta sketch: deserialize compact estimation from java", "[theta_sket
...
+  REQUIRE(sketch.get_theta() == Approx(update_sketch.get_theta()).margin(1e-10));
+  REQUIRE(sketch.get_estimate() == Approx(update_sketch.get_estimate()).margin(1e-10));
+  REQUIRE(sketch.get_lower_bound(1) == Approx(update_sketch.get_lower_bound(1)).margin(1e-10));
+  REQUIRE(sketch.get_upper_bound(1) == Approx(update_sketch.get_upper_bound(1)).margin(1e-10));
+  REQUIRE(sketch.get_lower_bound(2) == Approx(update_sketch.get_lower_bound(2)).margin(1e-10));
+  REQUIRE(sketch.get_upper_bound(2) == Approx(update_sketch.get_upper_bound(2)).margin(1e-10));
+  REQUIRE(sketch.get_lower_bound(3) == Approx(update_sketch.get_lower_bound(3)).margin(1e-10));
+  REQUIRE(sketch.get_upper_bound(3) == Approx(update_sketch.get_upper_bound(3)).margin(1e-10));
+  compact_theta_sketch compact_sketch = update_sketch.compact();
+  // the sketches are ordered, so the iteration sequence must match exactly
```
AlexanderSaydakov, line 275: this iteration relies on ordering, but there is a note above that v1 sketch may not be ordered

theta/test/theta_sketch_test.cpp
```diff
@@ -214,6 +244,78 @@ TEST_CASE("theta sketch: deserialize compact estimation from java", "[theta_sket
...
```
AlexanderSaydakov, line 311: same here. ordering check was disabled above, but this loop relies on the same order

theta/test/theta_sketch_test.cpp
```diff
@@ -296,4 +398,234 @@ TEST_CASE("theta sketch: conversion constructor and wrapped compact", "[theta_sk
...
 }
 
+TEST_CASE("theta sketch: wrap compact empty from java", "[theta_sketch]") {
+  std::ifstream is;
+  is.exceptions(std::ios::failbit | std::ios::badbit);
+  is.open(inputPath + "theta_compact_empty_from_java.sk", std::ios::binary | std::ios::ate);
+
+  std::vector<uint8_t> buf;
+  if(is) {
+      auto size = is.tellg();
```
AlexanderSaydakov, line 408: I would think that is.seekg(0, is.end) is needed first to obtain the file size with is.tellg()

theta/test/theta_sketch_test.cpp
```diff
@@ -296,4 +398,234 @@ TEST_CASE("theta sketch: conversion constructor and wrapped compact", "[theta_sk
...
+}
+
+TEST_CASE("theta sketch: wrap compact v1 empty from java", "[theta_sketch]") {
+  std::ifstream is;
+  is.exceptions(std::ios::failbit | std::ios::badbit);
+  is.open(inputPath + "theta_compact_empty_from_java_v1.sk", std::ios::binary | std::ios::ate);
+
+  std::vector<uint8_t> buf;
+  if(is) {
+      auto size = is.tellg();
```
AlexanderSaydakov, line 432: same here. I don't think this works as intended

theta/test/theta_sketch_test.cpp
```diff
@@ -296,4 +398,234 @@ TEST_CASE("theta sketch: conversion constructor and wrapped compact", "[theta_sk
...
+}
+
+TEST_CASE("theta sketch: wrap compact v2 empty from java", "[theta_sketch]") {
+  std::ifstream is;
+  is.exceptions(std::ios::failbit | std::ios::badbit);
+  is.open(inputPath + "theta_compact_empty_from_java_v2.sk", std::ios::binary | std::ios::ate);
+
+  std::vector<uint8_t> buf;
+  if(is) {
+      auto size = is.tellg();
```
AlexanderSaydakov, line 456: and here

theta/test/theta_sketch_test.cpp
```diff
@@ -296,4 +398,234 @@ TEST_CASE("theta sketch: conversion constructor and wrapped compact", "[theta_sk
...
+  REQUIRE(sketch.get_upper_bound(1) == 0.0);
+}
+
+TEST_CASE("theta sketch: wrap single item from java", "[theta_sketch]") {
+  std::ifstream is;
+  is.exceptions(std::ios::failbit | std::ios::badbit);
+  is.open(inputPath + "theta_compact_single_item_from_java.sk", std::ios::binary | std::ios::ate);
+  std::vector<uint8_t> buf;
+  if(is) {
+      auto size = is.tellg();
```
AlexanderSaydakov, line 479: and here, and more. I am not going to mark every instance

theta/include/compact_theta_sketch_parser_impl.hpp
```diff
@@ -29,32 +29,97 @@ template<bool dummy>
...
-  const uint32_t num_entries = reinterpret_cast<const uint32_t*>(ptr)[COMPACT_SKETCH_NUM_ENTRIES_U32];
-  const size_t entries_start_u64 = has_theta ? COMPACT_SKETCH_ENTRIES_ESTIMATION_U64 : COMPACT_SKETCH_ENTRIES_EXACT_U64;
-  const uint64_t* entries = reinterpret_cast<const uint64_t*>(ptr) + entries_start_u64;
-  const size_t expected_size_bytes = (entries_start_u64 + num_entries) * sizeof(uint64_t);
-  if (size < expected_size_bytes) {
-    throw std::invalid_argument(std::to_string(expected_size_bytes) + " bytes expected, actual " + std::to_string(size)
-        + (dump_on_error ? (", sketch dump: " + hex_dump(reinterpret_cast<const uint8_t*>(ptr), size)) : ""));
+  default:
+      // this should always fail since the valid cases are handled above
+      checker<true>::check_serial_version(reinterpret_cast<const uint8_t*>(ptr)[COMPACT_SKETCH_SERIAL_VERSION_BYTE], COMPACT_SKETCH_SERIAL_VERSION);
```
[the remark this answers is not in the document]
will-lauer: I left the check in place because it generates the better error message. Unfortunately, the compiler couldn't detect that the default case would always throw an exception in that case, causing a warning that one path (the default case) did not return a value. I added the throw, which should never be reached, to get rid of the warning.

theta/test/theta_sketch_test.cpp
will-lauer: I wasn't sure whether v1 sketches were guaranteed to be ordered or not. They definitely don't store an order flag, so I put the comment about ordering in. Whether or not the order flag is set, the data in the serialized sketch that we are reading from _is_ ordered, as I just look the v3 sketch and modified the preamble to conform with v1 serialization, so I think the test still holds, but I should probably change the comment.

theta/test/theta_sketch_test.cpp
will-lauer: Nope, seekg isn't needed, because the file was opened with `std:iso:ate`, which seeks automatically to the end of the file when it is opened. Since we start at the end, all we need is `tellg` to determine where we are, and then a `seekg` to get back to the beginning.

theta/test/theta_sketch_test.cpp
will-lauer: Checked with Lee, and v1 sketches are guaranteed to be ordered, so I've updated the code to set that flag, and to test that it is set in the unit tests

theta/include/compact_theta_sketch_parser_impl.hpp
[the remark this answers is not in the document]
AlexanderSaydakov: I don't quite like this, but not a problem. Perhaps can be addressed later. This also reminds me that I wanted to call this parser from the regular deserialization to avoid code duplication, but never got around to it.

theta/test/theta_sketch_test.cpp
AlexanderSaydakov: Right, I missed the "ate" flag. No problem then.

theta/test/theta_sketch_test.cpp
AlexanderSaydakov: ok

AlexanderSaydakov (approved):

PR #255, opened 2021-12-21
Theta: common code for wrap and deserialize

AlexanderSaydakov:
Reuse parser in deserialize from bytes.
This should also add support of serial versions 1 and 2, which was missing in deserialize from bytes.


jmalkin (approved):
Only reviewed the changes. I hadn't seen the compact deserialization before, but only focused on the changes here.

PR #258, opened 2022-01-11
KLL inclusive rank

AlexanderSaydakov:
This is to support inclusive get_rank(), get_PMF() and get_CDF() inspired by the implementation of REQ sketch.
Support of inclusive get_quantile() and get_quantiles() will be done separately by reusing the quantiles calculator from the REQ sketch.


jmalkin (approved):
Should probably add an explicit python test but not gonna make it a blocker

PR #259, opened 2022-01-31
Quantile sketch sorted view

AlexanderSaydakov:


common/include/quantile_sketch_sorted_view.hpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+  };
+
+  struct compare_pairs_by_second {
+    bool operator()(const Entry& a, const Entry& b) {
+      return a.second < b.second;
+    }
+  };
+
+  template<typename TT = T, typename std::enable_if<std::is_integral<TT>::value, int>::type = 0>
```
jmalkin, line 72: should this be `is_arithmetic`? i think we want it to work for floats and doubles, too?

jmalkin: same question for the next 3 declarations

common/include/quantile_sketch_sorted_view.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace datasketches {
+
+template<
+  typename T,
+  typename Comparator, // strict weak ordering function (see C++ named requirements: Compare)
+  typename Allocator
+>
+class quantile_sketch_sorted_view {
+public:
+  using Entry = typename std::conditional<std::is_integral<T>::value, std::pair<T, uint64_t>, std::pair<const T*, uint64_t>>::type;
```
jmalkin, line 34: i think `is_arithmetic`

common/include/quantile_sketch_sorted_view.hpp
```diff
@@ -0,0 +1,121 @@
...
+  size_t size() const;
+
+  // makes sense only with cumulative weight
+  const T& get_quantile(double rank) const;
+
+private:
+  static inline const T& deref_helper(const T* t) { return *t; }
+  static inline const T& deref_helper(const T& t) { return t; }
+
+  template<typename C>
```
jmalkin, line 59: Is this different from the `Comparator` defined in the class template?

jmalkin (changes requested):
fairly confident we want: s/is_integral/is_arithmetic/g

and i think we have an unnecessary comparator template argument

please correct me if i'm wrong on either point though!

common/include/quantile_sketch_sorted_view.hpp
AlexanderSaydakov: Yes, arithmetic

common/include/quantile_sketch_sorted_view.hpp
AlexanderSaydakov: yes

kll/include/kll_sketch.hpp
```diff
@@ -584,7 +591,7 @@ class kll_sketch {
 
     // implementations for all other types
     template<typename TT = T, typename std::enable_if<!std::is_floating_point<TT>::value, int>::type = 0>
-    static TT get_invalid_value() {
+    static const TT& get_invalid_value() {
       throw std::runtime_error("getting quantiles from empty sketch is not supported for this type of values");
```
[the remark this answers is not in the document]
AlexanderSaydakov: OK

common/include/quantile_sketch_sorted_view.hpp
AlexanderSaydakov: you are right, unnecessary template

jmalkin (approved):
I don't remember the original API right now  so I can't say for sure if this breaks anything? I think it's "kind of no, but edge cases like python's usage might break"?

either way, approving.

PR #261, opened 2022-02-04
Upload artifacts to pypi

richard-rogers:


jmalkin:
There are pypi packages provided by the datasketches team at this point. Why does whylabs need to continue using the datasketches trademark with their own name at this point?

PR #264, opened 2022-02-23
Reduced code duplication in KLL

AlexanderSaydakov:


kll/include/kll_sketch_impl.hpp
```diff
@@ -196,38 +191,14 @@ void kll_sketch<T, C, S, A>::merge(const kll_sketch& other) {
   const uint64_t final_n = n_ + other.n_;
   for (uint32_t i = other.levels_[0]; i < other.levels_[1]; i++) {
     const uint32_t index = internal_update();
-    new (&items_[index]) T(other.items_[i]);
+    new (&items_[index]) T(conditional_forward<FwdSk>(other.items_[i]));
```
jmalkin, line 194: should we be using `conditional_forward` for min and max values on lines 185 and 186, too? The old verionused `std::move` for those values in the now-removed code.

jmalkin (commented):
One potential change, so will wait on approving until we have a decision there.

kll/include/kll_sketch_impl.hpp
AlexanderSaydakov: right. that was an oversight. thank you

jmalkin (approved):
👍

Issue #262, opened 2022-02-12
datasketches does not work with python 3.9


MaheshGPai:
Datasketches module 3.2.0.1 seems to have been compiled with python 3.10 and does not work with python 3.9

> root@907f428e7dc2:/temp# pip install datasketches --target .
Collecting datasketches
  Downloading datasketches-3.2.0.1-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl (474 kB)
     |████████████████████████████████| 474 kB 658 kB/s 
Collecting numpy
  Downloading numpy-1.22.2-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.8 MB)
     |████████████████████████████████| 16.8 MB 7.3 MB/s 
Installing collected packages: numpy, datasketches
Successfully installed datasketches-3.2.0.1 numpy-1.22.2
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
WARNING: You are using pip version 21.2.4; however, version 22.0.3 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.

> root@907f428e7dc2:/temp# ls -lrt
total 1056
drwxr-xr-x  2 root root    4096 Feb 12 06:16 numpy.libs
drwxr-xr-x 19 root root    4096 Feb 12 06:16 numpy
drwxr-xr-x  2 root root    4096 Feb 12 06:16 bin
drwxr-xr-x  2 root root    4096 Feb 12 06:16 numpy-1.22.2.dist-info
-rwxr-xr-x  1 root root 1049008 Feb 12 06:16 datasketches.cpython-310-x86_64-linux-gnu.so
drwxr-xr-x  3 root root    4096 Feb 12 06:16 tests
drwxr-xr-x  3 root root    4096 Feb 12 06:16 src
drwxr-xr-x  2 root root    4096 Feb 12 06:16 datasketches-3.2.0.1.dist-info

> root@907f428e7dc2:/temp# python
Python 3.9.10 (main, Jan 26 2022, 20:56:53) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import datasketches
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'datasketches'

[...]

PR #272, opened 2022-05-04
Fix wheels

jmalkin:
This should address the issue in #262 , whcih ... I guess is some combination of a bug in cmake's FindPython and an ideological thing for manylinux?

The easiest solution is to require cmake 3.18 or higher, which unfortunately is just a tiny bit too new to be widely available on still-supported LTS platforms.


jmalkin:
I manually checked a number of manylinux wheels to confirm that the .so does not indicate python 3.10

will-lauer (approved):

PR #274, opened 2022-05-07
Common Kolmogorov-Smirnov test

AlexanderSaydakov:
Should work for both classic quantiles and KLL


common/include/kolmogorov_smirnov_impl.hpp
```diff
@@ -24,7 +24,7 @@ namespace datasketches {
 
 template<typename Sketch>
 double kolmogorov_smirnov::delta(const Sketch& sketch1, const Sketch& sketch2) {
-  using Comparator = typename Sketch::comparator;
+  auto comparator = sketch1.get_comparator(); // assuming the same comparator in sketch2
```
jmalkin, line 27: should we care if sketch1 and sketch2 have different comparators?

quantiles/include/quantiles_sketch_impl.hpp
```diff
@@ -262,7 +262,7 @@ void quantiles_sketch<T, C, A>::serialize(std::ostream& os, const SerDe& serde)
 template<typename T, typename C, typename A>
 template<typename SerDe>
 auto quantiles_sketch<T, C, A>::serialize(unsigned header_size_bytes, const SerDe& serde) const -> vector_bytes {
-  const size_t size = get_serialized_size_bytes() + header_size_bytes;
+  const size_t size = get_serialized_size_bytes(serde) + header_size_bytes;
```
jmalkin, line 265: good catch.. thx

quantiles/include/quantiles_sketch.hpp
```diff
@@ -146,14 +146,13 @@ namespace quantiles_constants {
...
-          typename Comparator = std::less<T>,
+          typename Comparator = std::less<T>, // strict weak ordering function (see C++ named requirements: Compare)
           typename Allocator = std::allocator<T>>
 class quantiles_sketch {
 public:
-  using C = Comparator;
-  using A = Allocator;
-  using AllocDouble = typename std::allocator_traits<Allocator>::template rebind_alloc<double>;
-  using vector_double = std::vector<double, AllocDouble>;
+  using value_type = T;
```
jmalkin, line 153: i don't think this is actually used, but i guess it gives us a consistent API and makes it easy to access

jmalkin (approved):

common/include/kolmogorov_smirnov_impl.hpp
AlexanderSaydakov: right now they are passed as a type, not instance. sketch1 and sketch2 have the same type. potentially we could have two template parameters to support sketches of different types. I am not sure we need this

quantiles/include/quantiles_sketch.hpp
AlexanderSaydakov: I tied to follow some practices of STL. they have many type aliases like this for users

PR #276, opened 2022-05-12
KLL cleanup

AlexanderSaydakov:
&array[index] can have undefined behavior if index == size to obtain an equivalent of the end iterator


jmalkin (approved):
👍

PR #281, opened 2022-06-03
KLL converting constructor

AlexanderSaydakov:


kll/include/kll_sketch.hpp
```diff
@@ -469,8 +491,8 @@ class kll_sketch {
     /**
      * This method deserializes a sketch from a given stream.
      * @param is input stream
-     * @param instance of a SerDe
-     * @param instance of an Allocator
+     * @param serge instance of a SerDe
```
jmalkin, line 494: type in serde

kll/include/kll_sketch_impl.hpp
```diff
@@ -147,6 +148,50 @@ kll_sketch<T, C, S, A>::~kll_sketch() {
...
+  levels_[1] = items_size_;
+
+  if (!other.is_empty()) {
+    min_value_ = new (allocator_.allocate(1)) T(other.get_min_value());
+    max_value_ = new (allocator_.allocate(1)) T(other.get_max_value());
+    size_t index = levels_[0];
+    for (auto pair: other) {
+      new (&items_[index]) T(pair.first);
+      const uint8_t level = count_trailing_zeros_in_u64(pair.second);
+      if (level == num_levels_) {
```
jmalkin, line 182: are you guaranteed that the iterator will never give level > num_levels_? I think this should probably be >=, but that may be a valid assumption here

jmalkin (commented):
1 typo and 1 suspicious line -- i think the latter either needs a fix or should have a comment to specify that the implicit assumption is ok

jmalkin (approved):
👍

PR #283, opened 2022-06-03
added missing serde parameter

AlexanderSaydakov:


jmalkin (approved):
good catch.

it'd be nice to have tests cover all these things, but that also starts making the testing increasingly difficult to maintain.

PR #289, opened 2022-06-30
Added the extra lower bound functions and test

c-dickens:
Added extra lower bound functionality for consistency with [java api](https://github.com/apache/datasketches-java/blob/7aed20ac52ba3bb18f3a427fac4edd94e9f8bf30/src/main/java/org/apache/datasketches/tuple/Sketch.java#L107) and likewise for upper bound.

Expanded unit testing framework to account for these changes as well.


tuple/include/tuple_sketch_impl.hpp
```diff
@@ -39,16 +39,30 @@ double tuple_sketch<S, A>::get_estimate() const {
   return get_num_retained() / get_theta();
 }
 
+template<typename S, typename A>
+double tuple_sketch<S, A>::get_lower_bound(uint8_t num_std_devs, uint32_t num_subset_entries) const {
+    if (is_empty()) num_subset_entries = 0 ; // When the sketch is empty we don't keep information for >0 identifiers.
```
AlexanderSaydakov, line 44: I don't think this line is needed. if sketch is empty, get_num_retained() == 0

c-dickens: I was originally unsure but I think that it is necessary.  Suppose the line is _not_ included and the user calls the function on an empty sketch but erroneously includes a `num_subset_entries = x` for an x > 0.  In this case, the sketch is empty, so the function should return 0.  However, because the sketch is empty and not in estimation mode, we move to line 46 and return `num_subset_entries = x` rather than zero.

AlexanderSaydakov: num_subset_entries will be set to 0 by line 45 since min(x, 0) == 0

c-dickens: yes, i agree now!

AlexanderSaydakov (approved):

PR #291, opened 2022-07-07
fix comparison bug since type uint8_t is always non-negative

dyf6372:
The type of `newShiftedVal` is uint8_t, so the condition `newShiftedVal < 0` is always false.


dyf6372:
> I am not sure why Python build on Windows failed

I'm also confused.

jmalkin:
It is unrelated to this change.

The Windows python build has an issue affecting all branches currently. It can possibly be resolved by fixing a version of some package but it's not yet clear which one. Alternatively there's a separate issue around what's included in our wheel and it's possible there's now a conflict? Hoping I'll be able to investigate next week.

If everything here works other than the Windows python CI tests we can consider it ok to merge as far as I'm concerned.

Issue #290, opened 2022-07-07
[python packaging] Top-level name problems


thatch:
Hi, looking at the 3.4.0 wheel for datasketches as released, the `*.dist-info/top_level.txt` file contains:

```
datasketches
src
tests
```

The top-level dir "src" should not be included at all; the one file it contains is not useful.
The top-level dir "tests" should be called something more unique (like "datasketches_tests") if you intend to package it in the wheel.  Alternatively just keep this in the sdist?

Both of these are likely to cause conflicts with other projects.

jmalkin:
Thanks for this, @thatch. We're not really python people so such suggestions are quite helpful.

We do expect to have some native python included at some point (trying to allow the use of generic python objects with container sketches, like kll or sampling, and that'll require user-specified classes for serialization or even comparators). I assume we'd want need to ensure they get included under `datasketches`?

thatch:
> I assume we'd want need to ensure they get included under datasketches?

Yep, the normal thing would be have this structure:

```
datasketches/__init__.py
datasketches/_datasketches.so
datasketches/kll.py
```

Right now you have `datasketches.so` and it is quite difficult to have both python and native parts of the same name.  You can have the `__init__.py` do `from ._datasketches import *` so all the names are available as a way of combining them.

jmalkin:
I'd experimented with the library part a bit a few months ago but we didn't have anything concrete to do with it then.

Right now we're trying to figure out why the Windows CI job started failing. Once that's done we'll be able to merge the linked PR (or maybe before if we determine it's sufficiently unrelated -- which really should be the case).

jmalkin:
This is now merged, along with a few other simplifications when revisiting the files. Thanks for pointing it out!

PR #293, opened 2022-07-14
remove pip as a build-system dependency in pyproject.toml

jamie256:
Remove pip as a build-system dependency in pyproject.toml to fix ci failure on Windows.

* Simplify pybind11Path.cmd to not require pip in favor of using pybind11's python method get_cmake_dir()
* incidental typo fix in README.md

The pyproject.toml build-system dependency on pip causes the following error when run on Windows (during the python tests) e.g.:
```
ERROR: To modify pip, please run the following command:
  D:\a\datasketches-cpp\datasketches-cpp\.tox\py3\Scripts\python.EXE -m pip install --ignore-installed --no-user --prefix C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-izfez3j3\overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- wheel setuptools >= 30.3.0 cmake >= 3.16 pip >= 10.0 pybind11[global] >= 2.6.0
```
For context on this error see discussion here: https://github.com/pypa/pip/issues/11120#issuecomment-1126669044

apache/datasketches-cpp appears to be hitting this error, so hopefully this change can resolve that:
https://github.com/apache/datasketches-cpp/runs/7328922009?check_suite_focus=true#step:8:39

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 requires = ["wheel",
             "setuptools >= 30.3.0",
             "cmake >= 3.16",
-            "pip >= 10.0",
             "pybind11[global] >= 2.6.0"]
 build-backend = "setuptools.build_meta"
 
--- a/python/README.md
+++ b/python/README.md
@@ -76,7 +76,7 @@ The only developer-specific instructions relate to running unit tests.
 
 ### Unit tests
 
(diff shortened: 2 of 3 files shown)


jamie256:
@jmalkin I posted this after seeing your comment here: https://github.com/apache/datasketches-cpp/issues/290#issuecomment-1181052167

jmalkin:
@jamie256 awesome -- thank you for this!  exactly what we needed.

jmalkin:
Realized I should probably have done an explicit review step. I did examine the changes prior to running the CI job.

PR #295, opened 2022-07-19
Remove catch.hpp and download pinned release verison from git instead

jmalkin:
Also, move to latest 2.x release.


common/test/CMakeLists.txt
```diff
@@ -33,7 +45,7 @@ target_include_directories(common_test
 
 target_sources(common_test
   INTERFACE
-    ${CMAKE_CURRENT_SOURCE_DIR}/catch.hpp  
+    #${CMAKE_CURRENT_SOURCE_DIR}/catch.hpp
```
will-lauer, line 48: Seems like you should get remove this instead of commenting it out

will-lauer (approved):

common/test/CMakeLists.txt
jmalkin: Thanks, removed

jmalkin:
One comment with a one-line fix to remove a commented-out line. Not gonna make anyone re-review that.

PR #297, opened 2022-07-26
Add python serde support for varopt sketches

jmalkin:
The immediate result is for varopt, but having this working should be an important part of adding tuple support.

I'm least confident in the error recovery during deserialization. I think it's most likely that Python will throw an error which propagates through the c++ code. In this case I'm intercepting it to try to decrement reference counts for items we've already allocated but it's hard to tell if that's actually preventing a leak or not. Happy for suggestions on how to clean up that code.

Also, fixed a size bug in varopt.

### DIFF
--- a/python/src/py_serde.cpp
+++ b/python/src/py_serde.cpp
@@ -17,21 +17,23 @@
  * under the License.
  */
 
-#include "py_serde.hpp"
+#include <cstring>
 #include "memory_operations.hpp"
+
+#include "py_serde.hpp"
 
 #include <pybind11/pybind11.h>
 
 namespace py = pybind11;
 
 void init_serde(py::module& m) {
-  py::class_<datasketches::py_object_serde, datasketches::PyObjectSerde /* <--- trampoline*/>(m, "PyObjectSerde")
+  py::class_<datasketches::py_object_serde, datasketches::PyObjectSerDe /* <--- trampoline*/>(m, "PyObjectSerDe")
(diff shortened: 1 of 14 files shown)


python/src/py_serde.cpp
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+#include <cstring>
+#include "memory_operations.hpp"
+
+#include "py_serde.hpp"
+
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
```
AlexanderSaydakov, line 27: pybind11 prefix is also used below

python/src/py_serde.cpp
```diff
@@ -0,0 +1,112 @@
...
+      py::tuple bytes_and_len;
+      try {
+        bytes_and_len = from_bytes(bytes, bytes_read);
+      } catch (py::error_already_set &e) {
+        failure = true;
+        error_from_python = true;
+        break;
+      }
+
+      // TODO: check bytes_and_len size is exactly 2?
```
AlexanderSaydakov, line 80: are you going to remove this? I am not sure the check is necessary.

AlexanderSaydakov (approved):
I don't think I completely understand how all this comes together, but looks fine to me

python/src/py_serde.cpp
jmalkin: Wasn't sure, so thx. Removed.

PR #298, opened 2022-08-10
Engagement test

c-dickens:
C++ version of Lee's [engagement test](https://github.com/apache/datasketches-java/blob/master/src/test/java/org/apache/datasketches/tuple/aninteger/EngagementTest.java)

### DIFF
--- a/tuple/test/CMakeLists.txt
+++ b/tuple/test/CMakeLists.txt
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-add_executable(tuple_test engagement_test.cpp)
+add_executable(tuple_test)
 
 target_link_libraries(tuple_test tuple common_test)
 
--- a/tuple/test/CMakeLists.txt
+++ b/tuple/test/CMakeLists.txt
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-add_executable(tuple_test)
+add_executable(tuple_test engagement_test.cpp)
 
 target_link_libraries(tuple_test tuple common_test)
 
@@ -45,4 +45,5 @@ target_sources(tuple_test
     tuple_a_not_b_test.cpp
     tuple_jaccard_similarity_test.cpp
     array_of_doubles_sketch_test.cpp
+    engagement_test.cpp
 )
--- a/tuple/test/engagement_test.cpp
+++ b/tuple/test/engagement_test.cpp
@@ -1,0 +1,338 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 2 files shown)


AlexanderSaydakov (approved):

tuple/test/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-add_executable(tuple_test)
+add_executable(tuple_test engagement_test.cpp)
```
jmalkin, line 18: I don't think we need to add this here since it's in target_sources below

tuple/test/engagement_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+    void update(T& summary, const T& update) const { }
+private:
+    T initial_value;
+};
+using always_one_tuple_sketch = datasketches::update_tuple_sketch<int, int, always_one_policy<int>> ;
+
+template<typename T>
+class update_sum_value_policy {
+public:
+    //update_sum_value_policy(const T& initial_value): initial_value(0) {}
```
jmalkin, line 54: ideally we'd remove comments like this when checking in a final version

tuple/test/engagement_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+                  << std::setw(12) << "Estimate"
+                  << std::setw(12) << "LB"
+                  << std::setw(12) << "UB"
+                  << std:: endl ;
+
+        for (uint64_t i = 0; i < num_days_arr.size(); i++) {
+            int visitors_at_days_visited = num_days_arr[i] ;
+            if(visitors_at_days_visited == 0){ continue; }
+            sum_visits += visitors_at_days_visited * i ;
+            // RUNS UP TO HERE 10/08/2022
```
jmalkin, line 260: ??

jmalkin (changes requested):
This doesn't quite match our existing style but that's not inherently problematic.

There are a few minor cleanup items.  I think I'd also prefer that `compute_engagement_histogram()` have a few `REQUIRES()` statements so it's testing something, and a way to disable printing to the terminal every time we run the test?

AlexanderSaydakov:
I think we should have a separate executable built with this example since it is not a unit test and prints stuff out

jmalkin:
> I think we should have a separate executable built with this example since it is not a unit test and prints stuff out

Doing this would also work, and would replace my suggestions (other than cleaning up commented out code)

tuple/test/CMakeLists.txt
c-dickens: Keeping this as is due to Alex's comments.

tuple/test/engagement_test.cpp
c-dickens: Comments removed.

jmalkin (approved):
👍

PR #299, opened 2022-08-17
Update  reverse_purge_hash_map copy constructor

why520it:
I think reverse_purge_hash_map copy constructor should do this here, the original result is not being able to copy all the values. Or remove the if and copy all the values directly, but it may affect some performance


why520it:
I think reverse_purge_hash_map copy constructor should do this here, the original result is not being able to copy all the values. Or remove the if and copy all the values directly, but it may affect some performance

AlexanderSaydakov:
thank you!

Issue #300, opened 2022-08-29
UndefinedBehaviorSanitizer failed, when serializing after using theta_a_not_b


pzhdfy:
We are integrating datasketches with Clickhouse.
In some test case, when serializing after using theta_a_not_b, we meet UndefinedBehavior(the binary built with -fsanitize=undefined). The stack.
```
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../contrib/datasketches-cpp/common/include/memory_operations.hpp:51:15 in
../contrib/datasketches-cpp/common/include/memory_operations.hpp:51:15: runtime error: null pointer passed as argument 2, which is declared to never be null
#0 0x148f9160 in datasketches::copy_to_mem(void const*, void*, unsigned long) build_docker/../contrib/datasketches-cpp/common/include/memory_operations.hpp:51:3
    #1 0x148f9160 in datasketches::compact_theta_sketch_alloc<std::__1::allocator<unsigned long> >::serialize(unsigned int) const build_docker/../contrib/datasketches-cpp/theta/include/theta_sketch_impl.hpp:391:12
    #2 0x148f8849 in DB::ThetaSketchData<unsigned long>::write(DB::WriteBuffer&) const build_docker/../src/AggregateFunctions/ThetaSketchData.h:158:49
    #3 0x24aa975a in DB::SerializationAggregateFunction::serializeBinaryBulk(DB::IColumn const&, DB::WriteBuffer&, unsigned long, unsigned long) const build_docker/../src/DataTypes/Serializations/SerializationAggregateFunction.cpp:73:19
    #4 0x26fd7b85 in DB::writeData(DB::ISerialization const&, COW<DB::IColumn>::immutable_ptr<DB::IColumn> const&, DB::WriteBuffer&, unsigned long, unsigned long) build_docker/../src/Formats/NativeWriter.cpp:63:19
    #5 0x26fd73f2 in DB::NativeWriter::write(DB::Block const&) build_docker/../src/Formats/NativeWriter.cpp:164:13
    #6 0x26f946d4 in DB::TCPHandler::sendData(DB::Block const&) build_docker/../src/Server/TCPHandler.cpp:1742:26
    #7 0x26f90829 in DB::TCPHandler::processOrdinaryQueryWithProcessors() build_docker/../src/Server/TCPHandler.cpp:748:21
    #8 0x26f821f3 in DB::TCPHandler::runImpl() build_docker/../src/Server/TCPHandler.cpp:368:17
[...]

PR #301, opened 2022-08-29
avoid needless call to memcpy with an empty entries_ array

jmalkin:
Based on #300 , we should avoid calling memcpy() when entries_ is empty to keep -fsanitize=undefiend happy. The sanitizer (reasonably) detected a copy of 0 bytes from nullptr. And while that should be completely harmless in this case, it's cleaner to avoid entirely.


AlexanderSaydakov (approved):

AlexanderSaydakov:
fixes #300

PR #302, opened 2022-08-30
No class level serde

AlexanderSaydakov:
removed serde from class template

### DIFF
--- a/sampling/include/var_opt_union_impl.hpp
+++ b/sampling/include/var_opt_union_impl.hpp
@@ -28,17 +28,17 @@
 
 namespace datasketches {
 
-template<typename T, typename S, typename A>
-var_opt_union<T,S,A>::var_opt_union(uint32_t max_k, const A& allocator) :
+template<typename T, typename A>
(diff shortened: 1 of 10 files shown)


sampling/include/var_opt_union_impl.hpp
```diff
@@ -201,7 +191,7 @@ var_opt_union<T,S,A> var_opt_union<T,S,A>::deserialize(const void* bytes, size_t
   bool is_empty = flags & EMPTY_FLAG_MASK;
 
   if (is_empty) {
-    return var_opt_union<T,S,A>(max_k);
+    return var_opt_union<T, A>(max_k);
```
jmalkin, line 194: probably don't need `<T,A>` here?

jmalkin (approved):
nice -- threw in a little excess notation cleanup, too, not just removing the template parameter.

sampling/include/var_opt_union_impl.hpp
AlexanderSaydakov: right

Issue #304, opened 2022-09-16
The Python package for ARM MacOS has an x86_64 datasketches.so in it


andrey-klochkov-liftoff:
When I install datasketches 3.5.0 with `pip` and the try to import it, I get the following on an ARM based MacBook:
```
>>> import datasketches
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: dlopen(/Users/username/venv/lib/python3.9/site-packages/datasketches.so, 0x0002): tried: '/Users/username/venv/lib/python3.9/site-packages/datasketches.so' (mach-o file, but is an incompatible architecture (have (x86_64), need (arm64e)))
```

I downloaded the file `datasketches-3.5.0-cp39-cp39-macosx_11_0_arm64.whl` from pypi.org, extracted `datasketches.so` from it and then checked the platform, and it's a wrong platform:
```
❯ file datasketches.so
datasketches.so: Mach-O 64-bit bundle x86_64
```

jmalkin:
Well, that certainly seems broken. Ugh. Thanks for flagging it.

I have an M1 Mac as of this week so I'll see if I can figure out what's happening -- and even be able to test it directly, finally!

andrey-klochkov-liftoff:
We just built the wheel from sources on an arm macbook, installed it and it worked fine. Seems something's off with the way those pypi packages were built, not the library code.

jmalkin:
Finally had a chance to look into this. Found an explanation from the same issue on other projects. The wheels are built via GitHub Actions, which apparently uses only Intel Macs as the actual task runners. The environment is mostly set to cross-compile -- but cmake is unaware of that when it configures the underlying C++ library.

That also explains why building a wheel locally works fine. Having a complete theory of the problem is good.

I'll try to test a patch later tonight.

jmalkin:
At long last, this has been resolved. Pypi now has v3.5.1, which fixes this. At least it worked when I tested it just now.

Apologies for the delay. The issue was very much not forgotten.

Issue #260, opened 2022-02-02
datasketches cannot be imported on Apple M1


davideanastasia:
I am currently experiencing a "wrong architecture" problem when I tried to use datasketches on my M1 macbook pro:

```
davide-anastasia@Davides-MBP ~ % mkdir datasketches
davide-anastasia@Davides-MBP ~ % cd datasketches
davide-anastasia@Davides-MBP datasketches % pipenv shell
Creating a virtualenv for this project...
Pipfile: /Users/davide-anastasia/datasketches/Pipfile
Using /opt/homebrew/bin/python3 (3.9.7) to create virtualenv...
⠸ Creating virtual environment...created virtual environment CPython3.9.7.final.0-64 in 169ms
  creator CPython3Posix(dest=/Users/davide-anastasia/.local/share/virtualenvs/datasketches-yWoxoifi, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/Users/davide-anastasia/Library/Application Support/virtualenv)
    added seed packages: pip==21.3.1, setuptools==60.2.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

✔ Successfully created virtual environment!
Virtualenv location: /Users/davide-anastasia/.local/share/virtualenvs/datasketches-yWoxoifi
Creating a Pipfile for this project...
Launching subshell in virtual environment...
 . /Users/davide-anastasia/.local/share/virtualenvs/datasketches-yWoxoifi/bin/activate
davide-anastasia@Davides-MBP datasketches %  . /Users/davide-anastasia/.local/share/virtualenvs/datasketches-yWoxoifi
/bin/activate
(datasketches) davide-anastasia@Davides-MBP datasketches % pipenv install datasketches
Installing datasketches...
Adding datasketches to Pipfile's [packages]...
✔ Installation Succeeded
Pipfile.lock not found, creating...
Locking [dev-packages] dependencies...
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success!
Updated Pipfile.lock (99e97b)!
Installing dependencies from Pipfile.lock (99e97b)...
[...]

PR #305, opened 2022-09-23
Update build_wheels

jmalkin:
Sets cmake environment variables to fix build problems for Apple Silicon (aka arm64) devices. Previously, the CI task knew the build for both Intel and ARM instruction sets, but cmake was detecting the platform (Intel) and always using that. As a result, the underlying C++ library was never actually cross-compiled for the target platform.

This should (finally) fix the python library for non-Intel Macs.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -62,12 +62,11 @@ jobs:
         with:
           python-version: '3.x'
            
-      #- name: Install Python dependencies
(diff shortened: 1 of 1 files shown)


jmalkin:
This should address #304 and even #260 -- I never got feedback on the older one, but I am now confident that my previous attempt did not actually fix the issue. I have successfully tested this fix with the Apple Silicon version, and also confirmed that the Intel version would not load on an M1 Mac.

AlexanderSaydakov (approved):

PR #309, opened 2022-10-05
serialization fix

AlexanderSaydakov:
similar to #303 for Tuple sketch


jmalkin (approved):
👍 
seemed super familiar since we did this on theta, and i guess forgot to port it here

PR #311, opened 2022-10-06
Universal sorted view

AlexanderSaydakov:
Sorted view to support both inclusive and exclusive modes at the same time.
Removed get_quantiles() method since sorted view is cached in a sketch now, and also can be obtained and queried many times.
More consistent terminology using a generic term "item"


python/src/vector_of_kll.cpp
```diff
@@ -276,31 +272,31 @@ py::array vector_of_kll_sketches<T,C,S>::get_num_retained() const {
 
 // Gets the minimum value of each sketch
 // TODO: allow subsets of sketches
-template<typename T, typename C, typename S>
-py::array vector_of_kll_sketches<T,C,S>::get_min_values() const {
+template<typename T, typename C>
+py::array vector_of_kll_sketches<T, C>::get_min_values() const {
```
jmalkin, line 276: do we want to stick with values() here or switch to items()? we can probably make the sketches hold py::objects eventually.

AlexanderSaydakov: missed this. will fix

AlexanderSaydakov: we instantiate vector_of_kll_sketches for int and float. perhaps we should keep "values" there

quantiles/include/quantiles_sketch.hpp
```diff
@@ -549,7 +530,7 @@ class quantiles_sketch {
   static uint32_t compute_retained_items(uint16_t k, uint64_t n);
   static uint32_t compute_base_buffer_items(uint16_t k, uint64_t n);
   static uint64_t compute_bit_pattern(uint16_t k, uint64_t n);
-  static uint32_t compute_valid_levels(uint64_t bit_pattern);
+  static uint32_t count_valid_levels(uint64_t bit_pattern);
```
jmalkin, line 533: not sure why the change here?

quantiles/include/quantiles_sketch.hpp
```diff
@@ -504,21 +481,25 @@ class quantiles_sketch {
   static const size_t DATA_START = 16;
 
   Allocator allocator_;
+  bool is_level_zero_sorted_;
```
jmalkin, line 484: if renaming this, it needs to be `is_base_buffer_sorted_` -- it is not levels_[0] but a different structure.

jmalkin (changes requested):
i think we need to rename `is_level_zero_sorted_` in the classic quantiles sketch to `is_base_buffer_sorted_` since level zero refers to a different vector.

quantiles/include/quantiles_sketch.hpp
AlexanderSaydakov: I felt that "compute" was too generic and vague. we really are counting them (counting set bits in the bit pattern)

jmalkin: everything else is `compute` but that's ok

jmalkin (approved):
👍

PR #322, opened 2023-01-03
Python wrapper improvement

AlexanderSaydakov:
added iterators, rearranged and simplified existing code

### DIFF
--- a/python/src/fi_wrapper.cpp
+++ b/python/src/fi_wrapper.cpp
@@ -17,62 +17,11 @@
  * under the License.
  */
 
+#include <pybind11/pybind11.h>
+
 #include "frequent_items_sketch.hpp"
 
-#include <pybind11/pybind11.h>
-#include <sstream>
-
 namespace py = pybind11;
-
-namespace datasketches {
-namespace python {
-
-template<typename T>
-frequent_items_sketch<T> fi_sketch_deserialize(py::bytes skBytes) {
-  std::string skStr = skBytes; // implicit cast  
-  return frequent_items_sketch<T>::deserialize(skStr.c_str(), skStr.length());
-}
-
-template<typename T>
-py::object fi_sketch_serialize(const frequent_items_sketch<T>& sk) {
-  auto serResult = sk.serialize();
-  return py::bytes((char*)serResult.data(), serResult.size());
-}
-
-// maybe possible to disambiguate the static vs method get_epsilon calls, but
-// this is easier for now
-template<typename T>
-double fi_sketch_get_generic_epsilon(uint8_t lg_max_map_size) {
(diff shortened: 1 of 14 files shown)


python/tests/vo_test.py
```diff
@@ -45,6 +45,13 @@ def test_vo_example(self):
     items = vo.get_samples()
     self.assertEqual(len(items), k)
 
+    count = 0
+    for tuple in vo:
+      sample = tuple[0]
+      weight = tuple[1]
```
jmalkin, line 51: worth summing weights and ensuring they equal the total?

jmalkin (commented):
There are still some C-style function pointer casts -- dunno if we wanna find them all first? And a few of the lambdas are complex enough that I'd have kept them as separate methods.

Also, I think we should wait until we fix the accidental push to the main branch before merging as it'll be a little cleaner to untangle that way.

python/src/fi_wrapper.cpp
```diff
@@ -105,15 +53,55 @@ void bind_fi_sketch(py::module &m, const char* name) {
...
-    .def_static("get_epsilon_for_lg_size", &dspy::fi_sketch_get_generic_epsilon<T>, py::arg("lg_max_map_size"),
-         "Returns the epsilon value used to compute a priori error for a given log2(max_map_size)")
-    .def_static("get_apriori_error", &frequent_items_sketch<T>::get_apriori_error, py::arg("lg_max_map_size"), py::arg("estimated_total_weight"),
-         "Returns the estimated a priori error given the max_map_size for the sketch and the estimated_total_stream_weight.")
-    .def("get_serialized_size_bytes", &dspy::fi_sketch_get_serialized_size_bytes<T>,
-         "Computes the size needed to serialize the current state of the sketch. This can be expensive since every item needs to be looked at.")
-    .def("serialize", &dspy::fi_sketch_serialize<T>, "Serializes the sketch into a bytes object")
-    .def_static("deserialize", &dspy::fi_sketch_deserialize<T>, "Reads a bytes object and returns the corresponding frequent_strings_sketch")
-    ;
+    .def(
```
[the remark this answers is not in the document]
AlexanderSaydakov: I thought that getting rid of that large section with namespaces would be good. Yes, this particular function is not trivial, but in my view it is not worse than having that large section just for this function.

python/tests/vo_test.py
AlexanderSaydakov: This seems to be a bit tricky since it is not exact. So I would suggest not test sketch correctness here.

jmalkin: Discussed offline. It's exact modulo machine precision with the arithmetic, but that's a fair point. It is reasonable to test that part only in C++.

jmalkin (approved):
LGTM

PR #324, opened 2023-01-09
Remove write-only variables

jbapple:
These variables were never read, which causes warning when compiling with clang 16

### DIFF
--- a/sampling/include/var_opt_sketch_impl.hpp
+++ b/sampling/include/var_opt_sketch_impl.hpp
@@ -754,10 +754,10 @@ string<A> var_opt_sketch<T, A>::items_to_string(bool print_gap) const {
   const uint32_t array_length = (n_ < k_ ? n_ : k_ + 1);
   for (uint32_t i = 0, display_idx = 0; i < array_length; ++i) {
     if (i == h_ && print_gap) {
-      os << i << ": GAP" << std::endl;
+      os << display_idx << ": GAP" << std::endl;
       ++display_idx;
     } else {
-      os << i << ": " << data_[i] << "\twt = ";
+      os << display_idx << ": " << data_[i] << "\twt = ";
       if (weights_[i] == -1.0) {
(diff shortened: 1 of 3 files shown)


sampling/include/var_opt_sketch_impl.hpp
```diff
@@ -752,18 +752,16 @@ string<A> var_opt_sketch<T, A>::items_to_string(bool print_gap) const {
   std::ostringstream os;
   os << "### Sketch Items" << std::endl;
   const uint32_t array_length = (n_ < k_ ? n_ : k_ + 1);
-  for (uint32_t i = 0, display_idx = 0; i < array_length; ++i) {
```
jmalkin, line 755: i think `os << i` should actually have been `os << display_idx` in the loop body

jmalkin:
Thanks for reporting this!

sampling/include/var_opt_sketch_impl.hpp
jbapple: Both places or just one?

jmalkin: Both

Not sure if you found this using this sketch or when compiling the library. The way it works, which isn't a necessary thing but helps simplify the algo, is that there's an array of size k+1, with a gap of size 1 between two portions of the sampling array that get slightly different treatment (heavy items vs items treated more like uniform reservoir sample). `to_string()` methods are largely just for debugging so providing an option to more easily track where that gap lives is useful -- but it changes indexing for items after the gap.

jbapple: Found it when compiling the library and running the included tests.

Changed to now use `display_idx` in the output.

jmalkin: thanks!

jmalkin (approved):
LGTM

jmalkin:
waiting for the CI job, then will merge

PR #325, opened 2023-01-09
Cpp countmin

c-dickens:
An initial attempt at cpp count min sketch.

Specific feedback that might be helpful:

- `get_sketch()` would this be better as an iterator over the sketch table, rather than returning the table explicitly?
- There is a danger of the hash seed addition overflowing.  See `count_min_impl.hpp` lines 24 & 115.  Is this problematic as it is passed as input to murmurhash?

### DIFF
--- a/count/include/count_min.hpp
+++ b/count/include/count_min.hpp
@@ -161,6 +161,14 @@ public:
   */
   void merge(const count_min_sketch<W> &other_sketch) ;
 
+  /**
+   * Returns true if this sketch is empty.
+   * A Count Min Sketch is defined to be empty iff weight == 0
+   * This can only ever happen if all items inserted to the sketch have weights that cancel each other out.
+   * @return empty flag
+   */
+  bool is_empty() const ;
+
   // Iterators
   using const_iterator = typename std::vector<W>::const_iterator ;
   const_iterator begin() const;
@@ -198,10 +206,10 @@ public:
    *
 
   0   ||    0   |    1   |    2   |    3   |    4   |    5   |    6   |    7   |
-      ||is_empty|00000001|00000001|00000001|xxxxxxxx|xxxxxxxx|xxxxxxxx|xxxxxxxx|
+      ||is_empty|ser__ver|familyId| flags  |xxxxxxxx|xxxxxxxx|xxxxxxxx|xxxxxxxx|
 
   1   ||    0   |    1   |    2   |    3   |    4   |    5   |    6   |    7   |
-      ||num_hash|----------- _num_buckets -----------|__seed__ __hash__|xxxxxxxx|
+      ||---------- _num_buckets -----------|num_hash|__seed__ __hash__|xxxxxxxx|
 
   2   ||    0   |    1   |    2   |    3   |    4   |    5   |    6   |    7   |
       ||---------------------------- total  weight ----------------------------|
@@ -213,6 +221,7 @@ public:
    *
    *
    */
+
   void serialize(std::ostream& os) const;
 
   /**
@@ -249,6 +258,24 @@ private:
   uint64_t _seed ;
   W _total_weight ;
   std::vector<uint64_t> hash_seeds ;
+
+  enum flags {IS_EMPTY};
+  static const uint8_t PREAMBLE_LONGS_SHORT = 2; // Empty -> need second byte for sketch parameters
+  static const uint8_t PREAMBLE_LONGS_FULL = 3; // Not empty -> need (at least) third byte for total weight.
+  static const uint8_t SERIAL_VERSION_1 = 1;
+  static const uint8_t FAMILY_ID = 1;
+  //static const size_t DATA_START = 24;
+  static const uint8_t NULL_8 = 0;
+  static const uint32_t NULL_32 = 0;
+
+  /**
+   * Throws an error if the header is not valid.
+   * @param preamble_longs
+   * @param serial_version
+   * @param flags_byte
+   */
+  static void check_header_validity(uint8_t preamble_longs, uint8_t serial_version, uint8_t family_id, uint8_t flags_byte);
+
 
 
 
--- a/count/include/count_min.hpp
+++ b/count/include/count_min.hpp
@@ -20,7 +20,7 @@ class count_min_sketch{
 public:
 
   /**
-  * Creates an instance of the sketch given parameters num_hashes, num_buckets and hash seed, `seed`.
+  * Creates an instance of the sketch given parameters _num_hashes, _num_buckets and hash seed, `seed`.
   * @param num_hashes : number of hash functions in the sketch. Equivalently the number of rows in the array
   * @param num_buckets : number of buckets that hash functions map into. Equivalently the number of columns in the array
   * @param seed for hash function
@@ -31,12 +31,12 @@ public:
   count_min_sketch(uint8_t num_hashes, uint32_t num_buckets, uint64_t seed = DEFAULT_SEED) ;
 
   /**
-  * @return configured num_hashes of this sketch
+  * @return configured _num_hashes of this sketch
   */
   uint16_t get_num_hashes() const;
 
   /**
-  * @return configured num_buckets of this sketch
+  * @return configured _num_buckets of this sketch
   */
   uint32_t get_num_buckets() const;
 
@@ -48,12 +48,12 @@ public:
   /**
    * @return epsilon : double
    * The maximum permissible error for any frequency estimate query.
-   * epsilon = ceil(e / num_buckets)
+   * epsilon = ceil(e / _num_buckets)
    */
    double get_relative_error() const;
 
   /**
-  * @return total_weight : typename W
+  * @return _total_weight : typename W
   * The total weight currently inserted into the stream.
   */
   W get_total_weight() const;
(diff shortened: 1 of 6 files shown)


count/include/count_min_impl.hpp
```diff
@@ -0,0 +1,234 @@
...
+namespace datasketches {
+
+template<typename T>
+count_min_sketch<T>::count_min_sketch(uint64_t num_hashes, uint64_t num_buckets, uint64_t seed):
+num_hashes(num_hashes),
+num_buckets(num_buckets),
+seed(seed){
+  sketch_length = num_hashes*num_buckets ;
+  sketch.resize(sketch_length) ;
+  assert(num_buckets >= 3); // need relative error epsilon at most 1.
```
AlexanderSaydakov, line 16: we have a policy to throw exceptions instead of asserts. std::invalid_argument in this case I suppose

count/include/count_min.hpp
```diff
@@ -0,0 +1,166 @@
...
+  * @param num_buckets : number of buckets that hash functions map into. Equivalently the number of columns in the array
+  * @param seed for hash function
+  *
+  * The template type T is the type of the vector that contains the weights, not the objects inserted into the sketch.
+   * The items inserted into the sketch can be arbitrary type, so long as they are hashable via murmurhash.
+   * Only update and estimate methods are added for uint64_t and string types.
+  */
+  count_min_sketch(uint64_t num_hashes, uint64_t num_buckets, uint64_t seed = DEFAULT_SEED) ;
+
+  std::vector<T> sketch ; // the array stored by the sketch
```
AlexanderSaydakov, line 41: should be private

count/include/count_min.hpp
```diff
@@ -0,0 +1,166 @@
+#ifndef COUNT_MIN_HPP_
+#define COUNT_MIN_HPP_
+
+#include <cstdint>
+#include <vector>
+#include <iterator>
+#include <algorithm>
+
```
AlexanderSaydakov, line 8: some of these includes are needed in the impl only

AlexanderSaydakov:
I think that overflowing seeed is fine. It still achieves seeding.

count/include/count_min.hpp
```diff
@@ -0,0 +1,166 @@
...
+   * @author Charlie Dickens
+   */
+
+template<typename T> class count_min_sketch ;
+
+template<typename T>
+class count_min_sketch{
+  static_assert(std::is_arithmetic<T>::value, "Arithmetic type expected");
+public:
+  uint64_t num_hashes, num_buckets, seed, sketch_length ;
```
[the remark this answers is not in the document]
c-dickens: - num_hashes has been changed to uint8.  The failure probability `delta = exp(-d)` and `exp(-255)` is on the order of 10^(-100) which will be small enough for applications.  On the other hand using a smaller integer type eg `exp(-15)` would rule out sketches with say 20 hash functions.  This is already a small probability of failure though.
- `num_buckets` has been changed to 32 bits by similar reasoning. 
- In future, we could make this flexible because 16 bit num_hashes and 32 bit num_buckets might still be larger than necessary for some applications.  Eg if the user can tolerate failure probability up to 1E-6, 1E-7, then 8 bit uints could be used for the number of hash functions, followed by a similar reduction in num_buckets if lower error can be tolerated.

count/include/count_min.hpp
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * void function which inserts an item of type std::string into the sketch
+   */
+  void update(const std::string& item, T weight) ;
+  void update(const std::string& item) ;
+
+  /*
+  * merges a separate count_min_sketch into this count_min_sketch.
+  */
+  void merge(count_min_sketch<T> &other_sketch) ;
```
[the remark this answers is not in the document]
c-dickens: Correct, resolved.

count/include/count_min.hpp
```diff
@@ -0,0 +1,166 @@
...
+  * @return total_weight : typename T
+  * The total weight currently inserted into the stream.
+  */
+  T get_total_weight() ;
+
+  /**
+  * @return vector of the sketch data structure
+  * Required for merging.
+  */
+  std::vector<T> get_sketch() ;
```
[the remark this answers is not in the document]
c-dickens: I have removed this function and updated with an iterator.

count/include/count_min.hpp
```diff
@@ -0,0 +1,189 @@
...
+  * @return configured seed of this sketch
+  */
+  uint64_t get_seed()  const;
+
+  /**
+   * @return epsilon : double
+   * The maximum permissible error for any frequency estimate query.
+   * epsilon = ceil(e / num_buckets)
+   */
+   double get_relative_error() ;
```
AlexanderSaydakov, line 53: this method should be const

count/include/count_min.hpp
```diff
@@ -0,0 +1,189 @@
...
+   */
+  W get_estimate(uint64_t item) ;
+
+   /**
+   * Specific get_estimate function for std::string type
+   * see generic get_estimate function
+   * @param item : std::string type
+   * @return an estimate of the item's frequency.
+   */
+  W get_estimate(const std::string& item) ;
```
AlexanderSaydakov, line 96: this method should be const

count/include/count_min.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+  * This is the generic estimate query function for any of the given datatypes.
+  * Query the sketch for the estimate of a given item.
+  * @param item : pointer to the data item to be query from the sketch.
+  * @param size : size_t
+  * @return the estimated frequency of the item denoted f_est satisfying
+  * f_true - relative_error*total_weight <= f_est <= f_true
+  */
+   W get_estimate(const void* item, size_t size) ;
```
AlexanderSaydakov, line 106: this method should be const, and get_*_bound() below too

count/include/count_min_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+namespace datasketches {
+
+template<typename W>
+count_min_sketch<W>::count_min_sketch(uint16_t num_hashes, uint32_t num_buckets, uint64_t seed):
+num_hashes(num_hashes),
+num_buckets(num_buckets),
+seed(seed){
+  total_weight = 0 ;
+  sketch_length = num_hashes*num_buckets ;
```
AlexanderSaydakov, line 15: this member variable does not seem to be necessary. it is easily derived and used once

count/include/count_min.hpp
c-dickens: Resolved.  Had to change `get_hashes(const void* item, size_t size)` to 
`get_hashes(const void* item, size_t size) const;` in order for all of these changes to const to compile correctly.

AlexanderSaydakov (approved):

count/include/count_min.hpp
```diff
@@ -0,0 +1,189 @@
...
+  /**
+  * Creates an instance of the sketch given parameters num_hashes, num_buckets and hash seed, `seed`.
+  * @param num_hashes : number of hash functions in the sketch. Equivalently the number of rows in the array
+  * @param num_buckets : number of buckets that hash functions map into. Equivalently the number of columns in the array
+  * @param seed for hash function
+  *
+  * The items inserted into the sketch can be arbitrary type, so long as they are hashable via murmurhash.
+  * Only update and estimate methods are added for uint64_t and string types.
+  */
+  count_min_sketch(uint16_t num_hashes, uint32_t num_buckets, uint64_t seed = DEFAULT_SEED) ;
```
c-dickens, line 31: The number of hashes does not need to be a 16-bit int.
The largest 5 bit int permits a failure probability of about  1/e^31 = 3E-12 which is plenty small enough.
the smallest datatype we can use is uint8_t so use that instead.

jmalkin:
Rerunning the specific component that failed with debug logs. I tried locally (although via mingw rather than a standard windows command prompt) and it was fine using the MS toolchain.

count/include/count_min_impl.hpp
```diff
@@ -0,0 +1,322 @@
...
+#include "MurmurHash3.h"
+#include <random>
+
+namespace datasketches {
+
+template<typename W>
+count_min_sketch<W>::count_min_sketch(uint8_t num_hashes, uint32_t num_buckets, uint64_t seed):
+_num_hashes(num_hashes),
+_num_buckets(num_buckets),
+_sketch_array(num_hashes * num_buckets, 0),
```
jmalkin, line 13: This will attempt to initialize a gigantic array, only to later decide that it's too big and throw an exception. We can instead do an initial check here:
`_sketch_array((num_hashes*num_buckets < 1<<30) ? num_hashes*num_buckets : 0, 0),`
which should let us re-enabld the unit test check and also avoid a potential overallocaiton issue for end-users.

jmalkin:
I think the Windows mystery is solved:
From [https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), Windows and Linux runners have a 7GB memory limit and that test was trying to allocated 8GB.

Not actually sure how it let Linux pass, but moving a check into the constructor initialization list will address this in a way that avoids issues for anyone using the code, too, which I think was just pushed.

AlexanderSaydakov (approved):

PR #329, opened 2023-01-11
Python tuple sketch support

jmalkin:
This adds tuple sketch support to the python library, and specifically provides an arbitrary python object as the summary, as well as custom python policies for set operations.

A few basic policy implementations are included. They both demonstrate examples of how to implement a custom policy as well as provide a few basic options that seem reasonably useful. (The always 1 policy perhaps less so.)

I think I should probably split the code in tuple_wrapper.cpp into a couple files, specifically pulling out the policy structs into their own header. But that would only ever be used from that one wrapper class so it's not obvious if that's actually useful?

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -42,6 +42,7 @@ target_link_libraries(python
     cpc
     fi
     theta
+    tuple
     sampling
     req
     quantiles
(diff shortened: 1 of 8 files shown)


jmalkin:
Oh! Just realized I have no tests.

Issue #333, opened 2023-01-19
.whl for Linux  ARM64 error


bllmo:
I have some trouble to install python `datasketches==4.0.0` on Linux  ARM64, I receive the following error when I run `pip3 install datasketches==4.0.0`

```
Building wheels for collected packages: datasketches
  Building wheel for datasketches (PEP 517) ... error
  ERROR: Command errored out with exit status 1:
   command: /usr/bin/python3 /usr/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py build_wheel /tmp/tmpr2gn81le
       cwd: /tmp/pip-install-r9ab7pna/datasketches
  Complete output (68 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-aarch64-cpython-37
  creating build/lib.linux-aarch64-cpython-37/datasketches
  copying python/datasketches/PySerDe.py -> build/lib.linux-aarch64-cpython-37/datasketches
  copying python/datasketches/__init__.py -> build/lib.linux-aarch64-cpython-37/datasketches
  running build_ext
  -- The CXX compiler identification is unknown
  CMake Error at CMakeLists.txt:25 (project):
    No CMAKE_CXX_COMPILER could be found.
  
    Tell CMake where to find the compiler by setting either the environment
    variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
    to the compiler, or to the compiler name if it is in the PATH.
  
  
  -- Configuring incomplete, errors occurred!
  See also "/tmp/pip-install-r9ab7pna/datasketches/build/temp.linux-aarch64-cpython-37/CMakeFiles/CMakeOutput.log".
  See also "/tmp/pip-install-r9ab7pna/datasketches/build/temp.linux-aarch64-cpython-37/CMakeFiles/CMakeError.log".
  Traceback (most recent call last):
    File "/usr/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/usr/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/usr/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 205, in build_wheel
      metadata_directory)
[...]

PR #334, opened 2023-01-21
Build linux arm64/aarch64 wheels

jmalkin:
QEMU seems to be the standard way to build wheels for arm64/aarch64 (same thing) with github actions. Cross compilation doesn't seem to be supported by cibuildwheel currently. The drawback of QEMU is that it's very, very slow.

Also removed a config step by putting the logic into a more appropriate step's env variables.

Tested a reduced set of platforms with a single python version and verified that the .so produced reports back as the right file type. I don't have an AWS instance to test this on, so waiting for the result of testing in  #333 .  If that works, this should be good to merge (and then maybe back-port into the 4.0.x branch).

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -3,10 +3,11 @@ name: Build Python Wheels
 on:
   # allow manual runs
   workflow_dispatch:
+  
   # run when we tag a release
-  release:
-    types:
-      - "created"
+  #release:
+  #  types:
+  #    - "created"
 
 env:
   BUILD_TYPE: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -3,10 +3,11 @@ name: Build Python Wheels
...
   # allow manual runs
   workflow_dispatch:
+  
   # run when we tag a release
-  release:
-    types:
-      - "created"
+  #release:
+  #  types:
+  #    - "created"
```
will-lauer, line 10: Why is this being commented out?

will-lauer (approved):

.github/workflows/build_wheels.yml
jmalkin: In practice, we've been including the wheels as part of a release vote. That means they need to be generated prior to creating a release tag, so anything automatically generated here wouldn't be used.

Maybe if someone downloaded the archive to extract the one they need manually as opposed to using pip install? But at that point it seems like they'd just as likely build a wheel locally.

PR #337, opened 2023-01-27
Fix doc strings in hll.hpp

jbapple:
While I'm here, include \<string\>, which is needed in some of the APIs.

### DIFF
--- a/hll/include/hll.hpp
+++ b/hll/include/hll.hpp
@@ -23,8 +23,9 @@
 #include "common_defs.hpp"
 #include "HllUtil.hpp"
 
-#include <memory>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


jmalkin:
Thank you.

Just curious, which compiler/IDE and settings complains about sorted header order? One of the longer-term plans is to increase our warnings settings and remove as many as we can. I think MSVC has the most, but also the most sensitive settings -- and is the most annoying to address since none of our primary devs use Windows.

jbapple:
Newer clang-format will fix header order for you.

jmalkin:
Mostly botch transfer from java, which I believe was me.

PR #338, opened 2023-01-27
Reduce likelihood of macro collision

jbapple:
. . . by removing one macro and prefacing the other with the logical module it's associated with.

### DIFF
--- a/common/include/MurmurHash3.h
+++ b/common/include/MurmurHash3.h
@@ -29,13 +29,13 @@ typedef unsigned char uint8_t;
 typedef unsigned int uint32_t;
 typedef unsigned __int64 uint64_t;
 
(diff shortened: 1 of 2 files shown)


jmalkin:
I feel like it might be worth doing this for the other constants, too?

Alternatively, would it make more sense to make this a namedspaced c++ thing?

jbapple:
> I feel like it might be worth doing this for the other macros, too?

Done.

> Alternatively, would it make more sense to make this a namedspaced c++ thing?

`ROTL64` and `CSA` (new in this commit) could be replaced by inline functions, which could be namespaced. The others cannot.

jmalkin:
I was thinking that we could alternatively restructure as a proper C++ class, but this certainly works for now and does improve naming safety. Thank you!

PR #340, opened 2023-01-28
use allocator properly in varopt union

jmalkin:
Tried to use a newer C++ standard and discovered varopt used deprecated allocator methods.  Those were relatively small, but I also discovered that the union didn't use the allocator properly at all.

### DIFF
--- a/sampling/include/var_opt_union_impl.hpp
+++ b/sampling/include/var_opt_union_impl.hpp
@@ -34,6 +34,7 @@ var_opt_union<T, A>::var_opt_union(uint32_t max_k, const A& allocator) :
   outer_tau_numer_(0.0),
   outer_tau_denom_(0),
   max_k_(max_k),
+  allocator_(allocator),
   gadget_(max_k, var_opt_sketch<T, A>::DEFAULT_RESIZE_FACTOR, true, allocator)
 {}
 
@@ -43,6 +44,7 @@ var_opt_union<T, A>::var_opt_union(const var_opt_union& other) :
   outer_tau_numer_(other.outer_tau_numer_),
   outer_tau_denom_(other.outer_tau_denom_),
   max_k_(other.max_k_),
+  allocator_(other.allocator_),
   gadget_(other.gadget_)
 {}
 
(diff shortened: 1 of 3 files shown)


sampling/include/var_opt_union_impl.hpp
```diff
@@ -508,9 +514,8 @@ void var_opt_union<T, A>::mark_moving_gadget_coercer(var_opt_sketch<T, A>& sk) c
   uint32_t result_r = 0;
   size_t next_r_pos = result_k; // = (result_k+1)-1, to fill R region from back to front
 
-  typedef typename std::allocator_traits<A>::template rebind_alloc<double> AllocDouble;
-  double* wts = AllocDouble().allocate(result_k + 1);
-  T* data     = A().allocate(result_k + 1);
+  double* wts = AllocDouble(allocator_).allocate(result_k + 1);
+  T* data     = A(allocator_).allocate(result_k + 1);
```
AlexanderSaydakov, line 518: why not just allocator_.allocate()?

sampling/include/var_opt_union_impl.hpp
```diff
@@ -554,11 +559,10 @@ void var_opt_union<T, A>::mark_moving_gadget_coercer(var_opt_sketch<T, A>& sk) c
...
   // clean up arrays in input sketch, replace with new values
-  typedef typename std::allocator_traits<A>::template rebind_alloc<bool> AllocBool;
-  AllocBool().deallocate(sk.marks_, sk.curr_items_alloc_);
-  AllocDouble().deallocate(sk.weights_, sk.curr_items_alloc_);
-  for (size_t i = 0; i < result_k; ++i) { A().destroy(sk.data_ + i); } // assumes everything in H region, no gap
-  A().deallocate(sk.data_, sk.curr_items_alloc_);
+  AllocBool(allocator_).deallocate(sk.marks_, sk.curr_items_alloc_);
+  AllocDouble(allocator_).deallocate(sk.weights_, sk.curr_items_alloc_);
+  for (size_t i = 0; i < result_k; ++i) { sk.data_[i].~T(); } // assumes everything in H region, no gap
+  A(allocator_).deallocate(sk.data_, sk.curr_items_alloc_);
```
AlexanderSaydakov, line 565: allocator_.deallocate() perhaps?

sampling/include/var_opt_union_impl.hpp
jmalkin: There were some places where it was unhappy doing that in a const function

AlexanderSaydakov: right

AlexanderSaydakov (approved):

sampling/include/var_opt_union_impl.hpp
jmalkin: Ok, just checked and the specific compiler error here is that it says there's no allocate() method defined in our test.

jmalkin: Sorry, read the whole error more carefully and it didn't match due to const issues.

sampling/include/var_opt_union_impl.hpp
jmalkin: Same as above -- it's because the function is const since the gadget is not changed.

PR #342, opened 2023-02-01
Remove inheritance from std::iterator

jmalkin:
Not entirely confident on all the difference types, and I punted on the pointer type to a std::pair<> we return since I don't think it'll matter with how we use things.

Anyway, I don't expect this to be the final version but this does at least seem to work with both gcc and clang (14) for a few different c++ standards (11 and 20 at least)

### DIFF
--- a/fi/include/reverse_purge_hash_map.hpp
+++ b/fi/include/reverse_purge_hash_map.hpp
@@ -91,8 +91,14 @@ private:
 
 // This iterator uses strides based on golden ratio to avoid clustering during merge
 template<typename K, typename V, typename H, typename E, typename A>
-class reverse_purge_hash_map<K, V, H, E, A>::iterator: public std::iterator<std::input_iterator_tag, K> {
+class reverse_purge_hash_map<K, V, H, E, A>::iterator {
 public:
+  using iterator_category = std::input_iterator_tag;
+  using value_type = std::pair<K&, V>;
+  using difference_type = std::ptrdiff_t;
+  using pointer = void;
+  using reference = value_type&;
+
   friend class reverse_purge_hash_map<K, V, H, E, A>;
   iterator& operator++() {
     ++count;
@@ -107,8 +113,8 @@ public:
   iterator operator++(int) { iterator tmp(*this); operator++(); return tmp; }
   bool operator==(const iterator& rhs) const { return count == rhs.count; }
   bool operator!=(const iterator& rhs) const { return count != rhs.count; }
-  const std::pair<K&, V> operator*() const {
-    return std::pair<K&, V>(map->keys_[index], map->values_[index]);
+  const value_type operator*() const {
+    return value_type(map->keys_[index], map->values_[index]);
   }
 private:
   static constexpr double GOLDEN_RATIO_RECIPROCAL = 0.6180339887498949; // = (sqrt(5) - 1) / 2
--- a/fi/include/reverse_purge_hash_map.hpp
+++ b/fi/include/reverse_purge_hash_map.hpp
@@ -97,7 +97,7 @@ public:
   using value_type = std::pair<K&, V>;
   using difference_type = void;
   using pointer = void;
-  using reference = value_type;
+  using reference = const value_type;
 
   friend class reverse_purge_hash_map<K, V, H, E, A>;
   iterator& operator++() {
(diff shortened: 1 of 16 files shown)


AlexanderSaydakov:
https://en.cppreference.com/w/cpp/named_req/InputIterator#Concept

"The reference type for an input iterator that is not also a LegacyForwardIterator does not have to be a reference type: dereferencing an input iterator may return a proxy object or value_type itself by value (as in the case of std::istreambuf_iterator)"

fi/include/reverse_purge_hash_map.hpp
```diff
@@ -91,8 +91,14 @@ class reverse_purge_hash_map {
 
 // This iterator uses strides based on golden ratio to avoid clustering during merge
 template<typename K, typename V, typename H, typename E, typename A>
-class reverse_purge_hash_map<K, V, H, E, A>::iterator: public std::iterator<std::input_iterator_tag, K> {
+class reverse_purge_hash_map<K, V, H, E, A>::iterator {
 public:
+  using iterator_category = std::input_iterator_tag;
+  using value_type = std::pair<K&, V>;
+  using difference_type = std::ptrdiff_t;
```
AlexanderSaydakov, line 98: I think this can be void in this case. I would be much more comfortable not specifying pointer difference for temporary values

fi/include/reverse_purge_hash_map.hpp
```diff
@@ -91,8 +91,14 @@ class reverse_purge_hash_map {
...
+  using pointer = void;
+  using reference = value_type&;
```
AlexanderSaydakov, line 100: I think we should make this = value_type

fi/include/reverse_purge_hash_map.hpp
```diff
@@ -107,8 +113,8 @@ class reverse_purge_hash_map<K, V, H, E, A>::iterator: public std::iterator<std:
   iterator operator++(int) { iterator tmp(*this); operator++(); return tmp; }
   bool operator==(const iterator& rhs) const { return count == rhs.count; }
   bool operator!=(const iterator& rhs) const { return count != rhs.count; }
-  const std::pair<K&, V> operator*() const {
-    return std::pair<K&, V>(map->keys_[index], map->values_[index]);
+  const value_type operator*() const {
```
AlexanderSaydakov, line 116: I think this should return reference type (same as value_type)

fi/include/reverse_purge_hash_map.hpp
```diff
@@ -107,8 +113,8 @@ class reverse_purge_hash_map<K, V, H, E, A>::iterator: public std::iterator<std:
...
+    return value_type(map->keys_[index], map->values_[index]);
```
AlexanderSaydakov, line 117: and instantiating value_type here is fine as implementation, I suppose

fi/include/reverse_purge_hash_map.hpp
jmalkin: Good point

fi/include/reverse_purge_hash_map.hpp
jmalkin: I don't see how the implementation can do anything else? It seems weird to implement as `reference(key, value)`

jmalkin:
Addressed the issues identified with the FI hash map iterator

jmalkin:
So is the suggestion that we should do that everywhere, including defining operator*() where it doesn't currently exist?

AlexanderSaydakov:
I think this approach with reference = value_type should be used everywhere we return a proxy object (all quantiles)

AlexanderSaydakov:
how can operator*() not exist?

jmalkin:
I think this is ready for review pass 2

AlexanderSaydakov:
> including defining operator*() where it doesn't currently exist

I you must be talking about operator->
I think I added it everywhere except the reverse purge hash map, which is for internal use

jmalkin:
> You must be talking about operator->

Yeah, that seems to be what I had in mind. It seems to have been missed for one of the sampling iterators so I'll see if I can get that in. Right now my local build has some sort of test failure so I clearly screwed up something.

jmalkin:
Now I think it _really_ is ready for re-review

AlexanderSaydakov (approved):

PR #343, opened 2023-02-10
Frequent items generic support

jmalkin:
I'm very much torn on whether defining `operator<<` on py::object is a good idea, even with it static to limit the scope. The sketch's `to_string()` method won't work without that, though. Even this way, it requires there to be a `__str__()` method defined for the python class to work usefully.

Unless there's a better way the only alternatives are that we write a new `to_string()` for generic objects calling the method directly, write a version that can't list the objects, or remove that method entirely.

But this feels rather hacky so I'm open to suggestions.

### DIFF
--- a/python/src/fi_wrapper.cpp
+++ b/python/src/fi_wrapper.cpp
@@ -98,7 +98,7 @@ void bind_fi_sketch(py::module &m, const char* name) {
         "Returns the estimated a priori error given the max_map_size for the sketch and the estimated_total_stream_weight."
     );
 
-    // serialization may need a caller-provided serde depending on teh sketch type, so
+    // serialization may need a caller-provided serde depending on the sketch type, so
     // we use a separate method to handle that appropriately based on type T.
     add_serialization(fi_class);
 }
@@ -157,14 +157,14 @@ void add_serialization(py::class_<datasketches::frequent_items_sketch<T, W, H, E
 
 // calls class __hash__ method
 struct py_hash_caller {
-  size_t operator()(const py::object& a) {
+  size_t operator()(const py::object& a) const {
     return py::hash(a);
   }
 };
 
 // calls class __eq__ method
 struct py_equal_caller {
-  bool operator()(const py::object& a, const py::object& b) {
+  bool operator()(const py::object& a, const py::object& b) const {
     return a.equal(b);
   }
 };
--- a/python/src/fi_wrapper.cpp
+++ b/python/src/fi_wrapper.cpp
@@ -17,45 +17,66 @@
  * under the License.
  */
 
+
+#include "py_serde.hpp"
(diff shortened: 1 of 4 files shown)


jmalkin:
Just realized I have no tests! That's a problem... But in the meantime, feel free to comment on whether this is even a viable approach.

python/src/fi_wrapper.cpp
```diff
@@ -73,37 +87,88 @@ void bind_fi_sketch(py::module &m, const char* name) {
...
         "get_apriori_error",
-        &frequent_items_sketch<T>::get_apriori_error,
+        &frequent_items_sketch<T, W, H, E>::get_apriori_error,
         py::arg("lg_max_map_size"), py::arg("estimated_total_weight"),
         "Returns the estimated a priori error given the max_map_size for the sketch and the estimated_total_stream_weight."
-    )
-    .def(
+    );
+
+    // serialization may need a caller-provided serde depending on teh sketch type, so
```
AlexanderSaydakov, line 101: typo

python/src/fi_wrapper.cpp
```diff
@@ -73,37 +87,88 @@ void bind_fi_sketch(py::module &m, const char* name) {
...
+        [](const std::string& bytes, py_object_serde& serde) {
+          return frequent_items_sketch<T, W, H, E>::deserialize(bytes.data(), bytes.size(), serde);
+        }, py::arg("bytes"), py::arg("serde"),
+        "Reads a bytes object using the provided serde and returns the corresponding frequent_strings_sketch."
+    );
+}
+
+// calls class __hash__ method
+struct py_hash_caller {
+  size_t operator()(const py::object& a) {
```
AlexanderSaydakov, line 160: this can be const, not sure if really matters

python/src/fi_wrapper.cpp
```diff
@@ -73,37 +87,88 @@ void bind_fi_sketch(py::module &m, const char* name) {
...
+    return py::hash(a);
+  }
+};
+
+// calls class __eq__ method
+struct py_equal_caller {
+  bool operator()(const py::object& a, const py::object& b) {
```
AlexanderSaydakov, line 167: and this can be const

python/src/fi_wrapper.cpp
jmalkin: fixed

python/src/fi_wrapper.cpp
jmalkin: right, this is done now

python/src/fi_wrapper.cpp
jmalkin: also done

AlexanderSaydakov (approved):

PR #346, opened 2023-02-23
HlLL get_rel_err() fix

AlexanderSaydakov:
check lgk range
handle > 12


jmalkin (approved):
Definitely better handling this way

PR #354, opened 2023-04-06
Count min python

jmalkin:
Adds python support for count_min, using double -- thinking maybe it should be int64_t weights by default instead? Or maybe unsigned if deletions aren't a thing?

### DIFF
--- a/python/src/count_wrapper.cpp
+++ b/python/src/count_wrapper.cpp
@@ -59,9 +59,9 @@ void bind_count_min_sketch(py::module &m, const char* name) {
          "Returns the maximum permissible error for any frequency estimate query")
     .def("get_total_weight", &count_min_sketch<W>::get_total_weight,
          "Returns the total weight currently inserted into the stream")
-    .def("update", static_cast<void (count_min_sketch<W>::*)(int64_t)>(&count_min_sketch<W>::update), py::arg("item"),
+    .def("update", static_cast<void (count_min_sketch<W>::*)(int64_t, W)>(&count_min_sketch<W>::update), py::arg("item"), py::arg("weight")=1.0,
          "Updates the sketch with the given 64-bit integer value")
-    .def("update", static_cast<void (count_min_sketch<W>::*)(const std::string&)>(&count_min_sketch<W>::update), py::arg("item"),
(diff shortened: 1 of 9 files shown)


python/src/count_wrapper.cpp
```diff
@@ -0,0 +1,101 @@
...
+void bind_count_min_sketch(py::module &m, const char* name) {
+  using namespace datasketches;
+
+  py::class_<count_min_sketch<W>>(m, name)
+    .def(py::init<uint8_t, uint32_t, uint64_t>(), py::arg("num_hashes"), py::arg("num_buckets"), py::arg("seed")=DEFAULT_SEED)
+    .def(py::init<const count_min_sketch<W>&>())
+    .def_static("suggest_num_buckets", &count_min_sketch<W>::suggest_num_buckets, py::arg("relative_error"),
+                "Suggests the number of buckets needed to achieve an accuracy within the provided "
+                "relative_error. For example, when relative_error = 0.05, the returned frequency estimates "
+                "satisfy the 'relative_error' guarantee that never overestimates the weights by may "
```
c-dickens, line 37: by maty -> but may

python/src/count_wrapper.cpp
```diff
@@ -0,0 +1,101 @@
...
+                "underestimate hte weights by 5% of the total weight in the sketch. "
```
c-dickens, line 38: hte -> the

python/tests/count_min_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # we can create another sketch with partial overlap
+    # and merge them
+    for i in range(int(n / 2), int(3 * n / 2)):
+      cm2.update(i, i)
+    cm.merge(cm2)
+
+    # and the estimated weight for the meerged values should now be
+    # at least 2x the value
+    self.assertGreaterEqual(cm.get_estimate(n), 2 * n)
```
c-dickens, line 67: Is it worth specifying that this only applies to the items in the overlap interval [n/2, n] ?
Items n+1, ... 3n / 2 - 1 have all been included only once with their weight.

c-dickens (approved):

c-dickens (commented):
I clicked approve having forgotten about the git build failure which will need resolving first.

PR #355, opened 2023-04-08
Density custom kernel

jmalkin:
* Adds support for providing a kernel instance in C++
* Adds support for native python kernels


jmalkin:
ugh.. this branch diverged a loooong time ago :(

Issue #359, opened 2023-04-15
Use non-colliding family id in count-min


jmalkin:
In order to have proper binary compatibility between C++ and Java and allow valid deserialization checks, we need to ensure we have a common family_id namespace. In Java it exists in a single enum class: https://github.com/apache/datasketches-java/blob/master/src/main/java/org/apache/datasketches/common/Family.java

The count-min sketch's family_id currently collides with Java's Alpha variant of the theta sketch: https://github.com/apache/datasketches-cpp/blob/master/count/include/count_min.hpp#L322
Let's assign count-min an id of 18, and then make a PR against the Java repo to reserve that value in the Family enum.

PR #360, opened 2023-04-17
Changed family id to 18

c-dickens:
Changed the family id to 18 as per [issue: # 359](https://github.com/apache/datasketches-cpp/issues/359)


c-dickens:
Actually, I don't think this is correct.  Changing the family id causes the header to change so the [acceptable parameters](https://github.com/apache/datasketches-cpp/blob/034749f10fdb16e94c634191b3d0757816a1df2b/count/include/count_min_impl.hpp#L487) in `check_header_validity` need to change accordingly.

jmalkin:
I think you identified the issue in `check_header_validity` but didn't get a chance to fix it yet?

jmalkin:
it worked when i used:
```
    case 138 : break; // !empty, ser_ver==1, family==18, preLongs=2;
    case 139 : break; // empty, ser_ver==1, family==18, preLongs=2;
    //case 170 : break ; // !empty, ser_ver==1, family==18, preLongs=3 ;
```

jmalkin (approved):

PR #362, opened 2023-04-18
Density serialization

jmalkin:
Adds density serialization

### DIFF
--- a/density/include/density_sketch_impl.hpp
+++ b/density/include/density_sketch_impl.hpp
@@ -39,6 +39,17 @@ levels_(1, Level(allocator), allocator)
 {}
 
 template<typename T, typename K, typename A>
+density_sketch<T, K, A>::density_sketch(uint16_t k, uint32_t dim, uint32_t num_retained, uint64_t n,
+                                        Levels&& levels, const K& kernel):
+kernel_(kernel),
+k_(k),
+dim_(dim),
+num_retained_(num_retained),
+n_(n),
+levels_(std::move(levels))
+{}
+
+template<typename T, typename K, typename A>
 uint16_t density_sketch<T, K, A>::get_k() const {
   return k_;
 }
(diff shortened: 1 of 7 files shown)


density/include/density_sketch_impl.hpp
```diff
@@ -146,6 +158,253 @@ void density_sketch<T, K, A>::compact_level(unsigned height) {
...
+  uint8_t flags_byte;
+  ptr += copy_from_mem(ptr, flags_byte);
+  uint16_t k;
+  ptr += copy_from_mem(ptr, k);
+  uint16_t unused;
+  ptr += copy_from_mem(ptr, unused);
+  uint32_t dim;
+  ptr += copy_from_mem(ptr, dim);
+
+  //check_k(k);
```
AlexanderSaydakov, line 329: do we need to check K here?

jmalkin: We don't check anything in the constructor so I dunno what the value range would be? It's uint16_t so it obviously can't be negative.  We can check that it's strictly positive?  It seems like there should be some minimum practical value given the need for compression, though.

AlexanderSaydakov: I am not sure we need to check, but I saw this commented-out line, which brings a question whether to remove it or implement.

jmalkin: My top choice would be to check. I'm not sure of a practical minimum though. Fairly sure 1 is not a good idea, and 0 should not be allowed. So maybe uncomment, check >1, and also add to the constructor?

AlexanderSaydakov: I would agree to that, if there are no other practical limits

density/include/density_sketch_impl.hpp
```diff
@@ -146,6 +162,259 @@ void density_sketch<T, K, A>::compact_level(unsigned height) {
...
+
+  if (is_empty())
+    return;
+
+  write(os, num_retained_);
+  write(os, n_);
+
+  // levels array -- uint32_t since a single level may be larger than k
+  size_t pt_size = sizeof(T) * dim_;
+  for (Level lvl : levels_) {
```
AlexanderSaydakov, line 208: shouldn't this be a (const) ref?

density/include/density_sketch_impl.hpp
```diff
@@ -146,6 +162,259 @@ void density_sketch<T, K, A>::compact_level(unsigned height) {
...
+    const uint32_t level_size = static_cast<uint32_t>(lvl.size());
+    write(os, level_size);
+    for (Vector pt : lvl) {
```
AlexanderSaydakov, line 211: and here const ref perhaps?

density/include/density_sketch_impl.hpp
```diff
@@ -146,6 +162,259 @@ void density_sketch<T, K, A>::compact_level(unsigned height) {
...
+  
+  if (is_empty())
+    return bytes;
+
+  ptr += copy_to_mem(num_retained_, ptr);
+  ptr += copy_to_mem(n_, ptr);
+
+  // levels array -- uint32_t since a single level may be larger than k
+  size_t pt_size = sizeof(T) * dim_;
+  for (Level lvl : levels_) {
```
AlexanderSaydakov, line 252: const ref

density/include/density_sketch_impl.hpp
```diff
@@ -146,6 +162,259 @@ void density_sketch<T, K, A>::compact_level(unsigned height) {
...
+    ptr += copy_to_mem(static_cast<uint32_t>(lvl.size()), ptr);
+    for (Vector pt : lvl) {
```
AlexanderSaydakov, line 254: const fer

AlexanderSaydakov (approved):

PR #363, opened 2023-04-19
Density sketch

jmalkin:
Don't know if we want to spend more time on characterization, but I believe the sketch itself is ready for inclusion in the library. The conflicts should just be reconciling the CMakeLists.txt files to play nicely with the count-min sketch.


jmalkin:
Yup, fixed CMakeLists.txt files (main and python) as well as the python files to import all the relevant pieces. And reconciliation was to just take the changes from both.

AlexanderSaydakov (approved):

PR #364, opened 2023-04-21
Hll delay kxq

jmalkin:
This changes HLL union to delay recomputing the KxQ and cur_min values until the union gadget needs to be made a consistent sketch (either get_result() or a direct get_estimate() will trigger that). This mirrors how HLL union operates in java.

Using characterization's hll-union-timing test on my linux desktop showed a steady-state merge speed improvement around 2.8x.

### DIFF
--- a/hll/include/Hll4Array-internal.hpp
+++ b/hll/include/Hll4Array-internal.hpp
@@ -52,6 +52,22 @@ Hll4Array<A>::Hll4Array(const Hll4Array<A>& that) :
 }
 
 template<typename A>
+Hll4Array<A>::Hll4Array(const HllArray<A>& other) :
+  HllArray<A>(other.getLgConfigK(), target_hll_type::HLL_4, other.isStartFullSize(), other.getAllocator()),
+  auxHashMap_(nullptr)
+{
+  const int numBytes = this->hll4ArrBytes(this->lgConfigK_);
+  this->hllByteArr_.resize(numBytes, 0);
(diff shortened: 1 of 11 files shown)


hll/include/Hll4Array-internal.hpp
```diff
@@ -51,6 +51,22 @@ Hll4Array<A>::Hll4Array(const Hll4Array<A>& that) :
...
+template<typename A>
+Hll4Array<A>::Hll4Array(const HllArray<A>& other) :
+  HllArray<A>(other.getLgConfigK(), target_hll_type::HLL_4, other.isStartFullSize(), other.getAllocator()),
+  auxHashMap_(nullptr)
+{
+  const int numBytes = this->hll4ArrBytes(this->lgConfigK_);
+  this->hllByteArr_.resize(numBytes, 0);
+  this->oooFlag_ = other.isOutOfOrderFlag();
+
+  for (const auto& coupon : other) { // all = false, so skip empty values
```
AlexanderSaydakov, line 63: coupon is a small integral type thing, so, perhaps, copying is better than reference, but, I guess, not very important

AlexanderSaydakov (approved):

hll/include/Hll4Array-internal.hpp
jmalkin: fixed anyway

AlexanderSaydakov (approved):

PR #371, opened 2023-05-03
remove std::iterator from wrapped_compact_theta_sketch_alloc

jmalkin:


jmalkin:
Should fix #370

AlexanderSaydakov:
> const uint64_t& operator*() const;
> const uint64_t* operator->() const;

these should return "reference" and "pointer"

theta/include/theta_sketch.hpp
```diff
@@ -471,8 +471,14 @@ class wrapped_compact_theta_sketch_alloc : public base_theta_sketch_alloc<Alloca
 };
 
 template<typename Allocator>
-class wrapped_compact_theta_sketch_alloc<Allocator>::const_iterator: public std::iterator<std::input_iterator_tag, uint64_t> {
+class wrapped_compact_theta_sketch_alloc<Allocator>::const_iterator {
 public:
+  using iterator_category = std::input_iterator_tag;
+  using value_type = uint64_t;
```
AlexanderSaydakov, line 477: const

theta/include/theta_sketch.hpp
```diff
@@ -471,8 +471,14 @@ class wrapped_compact_theta_sketch_alloc : public base_theta_sketch_alloc<Alloca
...
+  using difference_type = void;
+  using pointer = uint64_t*;
```
AlexanderSaydakov, line 479: const

theta/include/theta_sketch.hpp
```diff
@@ -471,8 +471,14 @@ class wrapped_compact_theta_sketch_alloc : public base_theta_sketch_alloc<Alloca
...
+  using reference = uint64_t;
```
AlexanderSaydakov, line 480: const

AlexanderSaydakov:
While we are at this, let's add operator->() to theta_update_sketch_base theta_iterator and theta_const_iterator

theta/include/theta_sketch.hpp
jmalkin: gcc complains (warning) about `const` here in the return type.

AlexanderSaydakov: right, let't not use const when returning by value

theta/include/theta_union_base_impl.hpp
```diff
@@ -42,7 +42,7 @@ void theta_union_base<EN, EK, P, S, CS, A>::update(SS&& sketch) {
   if (sketch.get_seed_hash() != compute_seed_hash(table_.seed_)) throw std::invalid_argument("seed hash mismatch");
   table_.is_empty_ = false;
   union_theta_ = std::min(union_theta_, sketch.get_theta64());
-  for (auto& entry: sketch) {
+  for (auto entry: sketch) {
```
AlexanderSaydakov, line 45: I think we need auto&& here too

AlexanderSaydakov (approved):

PR #379, opened 2023-06-29
documentation fixes

AlexanderSaydakov:


AlexanderSaydakov:
Besides doc fixes I made a couple of HLL operator= methods return ref instead of value, and removed a couple of update methods in count_min_sketch, which became unnecessary after adding default weight of 1. I believe these changes must be invisible for the user.

theta/include/theta_sketch.hpp
```diff
@@ -106,6 +107,7 @@ class base_theta_sketch_alloc {
   virtual void print_items(std::ostringstream& os) const = 0;
 };
 
+/// Base class for the Theta Sketch, a generalization of the Kth Minimum Value (KMV) sketch.
```
jmalkin, line 110: Is it a generalization of or a type of KMV sketch? Maybe sidestep and describe it as a type of KMV sketch for counting distinct items?

tuple/include/tuple_sketch.hpp
```diff
@@ -522,8 +558,7 @@ class compact_tuple_sketch: public tuple_sketch<Summary, Allocator> {
 
 };
 
-// builder
-
+// base builder
```
jmalkin, line 583: intentional to have only `//` rather than `///`?

jmalkin (approved):
A few minor comments but nothing obviously objectionable in the current state

theta/include/theta_sketch.hpp
AlexanderSaydakov: we have this on the web site: "Theta Sketches are a generalization of the well known Kth Minimum Value (KMV)1,2 sketches in that KMV sketches are a form of Theta Sketch, but not all Theta Sketches are KMV."

tuple/include/tuple_sketch.hpp
AlexanderSaydakov: I think we could expose this one as well

theta/include/theta_sketch.hpp
jmalkin: Ok, although a bit misleading in the context of C++ where we don't offer anything but KMV. That's fine though.

jmalkin (approved):
LGTM

PR #381, opened 2023-07-07
avoid marketplace action

AlexanderSaydakov:
Replaced the first marketplace doxygen action with simple commands.
The second marketplace action is still there.
Let's see if it works.


jmalkin:
Tested the workflow and it ran successfully. Yay!

jmalkin (approved):

PR #382, opened 2023-07-13
simple substitute for std::optional until we require C++17

AlexanderSaydakov:


jmalkin (approved):
Getting rid of the item deleter is awesome. I like it.

PR #387, opened 2023-07-21
test deserialize sketches from Java

AlexanderSaydakov:
KLL float and double for now


.github/workflows/serde_compat.yml
```diff
@@ -0,0 +1,27 @@
+name: Java SerDe Compatibility Test
+
+on:
+  push:
+    branches:
+      - master
```
jmalkin, line 6: do we want a manual trigger option for if we're trying to update serialization in a sketch, without needing to open a PR?

cpc/test/cpc_sketch_deserialize_from_java_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+#include <catch2/catch.hpp>
+#include <fstream>
+#include <cpc_sketch.hpp>
+
+namespace datasketches {
+
+static std::string testBinaryInputPath = std::string(TEST_BINARY_INPUT_PATH) + "../../java/";
```
jmalkin, line 26: This needs more description since it should be clear how to run it locally.

This works by running the java tests and just leaving files in the top level directory of that repo?

cpc/test/cpc_sketch_deserialize_from_java_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+TEST_CASE("cpc sketch", "[serde_compat]") {
+  unsigned n_arr[] = {0, 100, 200, 2000, 20000};
```
jmalkin, line 29: maybe not currently feasible but it'd be good not to need to coordinate hard-coded arrays between language repos if possible

jmalkin (approved):

.github/workflows/serde_compat.yml
AlexanderSaydakov: perhaps. I will add the manual trigger

cpc/test/cpc_sketch_deserialize_from_java_test.cpp
AlexanderSaydakov: right, I will add some comments

cpc/test/cpc_sketch_deserialize_from_java_test.cpp
AlexanderSaydakov: That would require some metadata saved next to the sketch binaries, some parser for it, and still some agreement about the format and supported features. Possible, but not now.

PR #390, opened 2023-08-07
Test serialization compatibility with Java

AlexanderSaydakov:
Added varopt and frequent items sketch tests, fixed IS_EMPTY flag mismatch with Java in frequent items sketch


jmalkin (approved):
Good catch on using `stoi()` and a custom comparator

PR #391, opened 2023-08-16
removed static tests replaced by dynamic ones

AlexanderSaydakov:


kll/test/kll_sketch_deserialize_from_java_test.cpp
```diff
@@ -75,4 +75,29 @@ TEST_CASE("kll double", "[serde_compat]") {
   }
 }
 
+// numbers are padded with leading spaces so that natural order works
```
jmalkin, line 78: this is automatic in the serde? there doesn't seem to be anything in the code that'd handle that?

jmalkin (approved):
Not entirely sure what's meant by one comment but not a blocker.

kll/test/kll_sketch_deserialize_from_java_test.cpp
AlexanderSaydakov: this is how it was done on the Java side during sketch updates presumably to avoid supplying a custom comparator

PR #392, opened 2023-08-26
Generate sketches for java

AlexanderSaydakov:


sampling/test/var_opt_union_serialize_for_java.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * under the License.
+ */
+
+#include <catch2/catch.hpp>
+#include <fstream>
+#include <var_opt_union.hpp>
+
+namespace datasketches {
+
+TEST_CASE("var opt union doulbe sampling", "[serialize_for_java]") {
```
jmalkin, line 26: typo: double

sampling/test/var_opt_union_serialize_for_java.cpp
```diff
@@ -0,0 +1,56 @@
...
+  const unsigned k_small = 16;
+  const unsigned k_max = 128;
+  const unsigned n1 = 32;
+  const unsigned n2 = 64;
+
+  // small k sketch, but sampling
+  var_opt_sketch<double> sketch1(k_small);
+  for (unsigned i = 0; i < n1; ++i) sketch1.update(i);
+  // heavy items have negative weights to allow a simple predicate to filter
```
jmalkin, line 35: s/weights/values/

theta/test/theta_sketch_serialize_for_java.cpp
```diff
@@ -0,0 +1,59 @@
...
+    REQUIRE_FALSE(sketch.is_empty());
+    REQUIRE(sketch.get_estimate() == Approx(n).margin(n * 0.03));
+    std::ofstream os("theta_compressed_n" + std::to_string(n) + "_cpp.sk", std::ios::binary);
+    sketch.compact().serialize(os);
+  }
+}
+
+TEST_CASE("theta sketch generate non-empty no entries", "[serialize_for_java]") {
+  auto sketch = update_theta_sketch::builder().set_p(0.01).build();
+  sketch.update(1);
```
jmalkin, line 52: We do this for AOD, too. It does depend on the hash function vs p=0.01. I don't think it's critical but it is conditioning a test on an an implicit assumption around the hash function.

jmalkin (approved):
A few minor things noted. Nothing serious, so no need to re-review after fixing comments.

theta/test/theta_sketch_serialize_for_java.cpp
AlexanderSaydakov: right, but because 1 always has the same hash, and the test passes, it will stay that way

fi/test/frequent_items_sketch_serialize_for_java.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+TEST_CASE("frequent strings sketch generate", "[serialize_for_java]") {
+  const unsigned n_arr[] = {0, 1, 10, 100, 1000, 10000, 100000, 1000000};
+  for (const unsigned n: n_arr) {
+    frequent_items_sketch<std::string> sketch(6);
+    for (unsigned i = 1; i <= n; ++i) sketch.update(std::to_string(i));
+    REQUIRE(sketch.is_empty() == (n == 0));
+    if (n > 10) {
+//      REQUIRE(sketch.get_maximum_error() > 0);
```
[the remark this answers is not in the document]
AlexanderSaydakov: forgot to uncomment after some experiment

PR #393, opened 2023-09-06
removed python stuff since it was moved to its own repo

AlexanderSaydakov:


jmalkin (approved):
byebye python... (and welcome python repo)

PR #399, opened 2023-10-23
more generic implementation of array_of_doubles_sketch

AlexanderSaydakov:
array_tuple_sketch , union, intersection, a_not_b with parameterized array type and policy


AlexanderSaydakov:
Motivation:
I found that array_of_doubles_* was too rigid: doubles only, no way to change update policy and union policy, and no way to substitute our custom array with std::vector
In this implementation all array_of_doubles_* types are just aliases with default policy (if applicable) and array<double>, but easy to have any type like std::vector<int> or something.
I am not sure if the name array_tuple_* is the best. Suggestions are welcome.

jmalkin (approved):

tuple/test/array_of_doubles_sketch_test.cpp
```diff
@@ -166,7 +164,7 @@ TEST_CASE("aod intersection: half overlap", "[tuple_sketch]") {
   auto update_sketch2 = update_array_of_doubles_sketch::builder().build();
   for (int i = 500; i < 1500; ++i) update_sketch2.update(i, a);
 
-  array_of_doubles_intersection<array_of_doubles_union_policy> intersection;
```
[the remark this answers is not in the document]
AlexanderSaydakov: yes, let's add a comment

jmalkin (approved):
Thx. I think this is easier to read for understanding behavior now.

tuple/include/array_tuple_union_impl.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace datasketches {
+
+template<typename Array, typename Policy, typename Allocator>
+array_tuple_union<Array, Policy, Allocator>::array_tuple_union(uint8_t lg_cur_size, uint8_t lg_nom_size, resize_factor rf, float p, uint64_t theta, uint64_t seed, const Policy& policy, const Allocator& allocator):
+Base(lg_cur_size, lg_nom_size, rf, p, theta, seed, policy, allocator)
+{}
+
+template<typename Array, typename Policy, typename Allocator>
+auto array_tuple_union<Array, Policy, Allocator>::get_result(bool ordered) const -> CompactSketch {
+  return CompactSketch(this->state_.get_policy().get_policy().get_num_values(), Base::get_result(ordered));
```
jmalkin, line 29: do we need to change this, too?

AlexanderSaydakov: yes, why not, while we are at this

jmalkin (approved):

jmalkin:
Merge away!

PR #402, opened 2023-10-28
Add install to CI job

jmalkin:


jmalkin:
Tested the workflow change in my own fork to avoid spamming the mailing list. Merging this into a branch I had created earlier to address the superficial issue.

This isn't a complete fix in that we don't check that all intended files are present in the install but it will at least check that anything we rename is properly updated.

PR #406, opened 2023-11-16
use global CXX_STANDARD setting

AlexanderSaydakov:


jmalkin (approved):
i'm not 100% sure if this would pose problems for people depending on a non-installed version. I'd ideally like a second opinion from @will-lauer (or anyone with more knowledge of cmake)

will-lauer:
Honestly, I don't see why this would impact non-installed versions, but I don't really understand what that command in cmake actually does. The only suggestion I have is to try it out. I suspect if you try building python with the new changes, you'll see whether they have a significant impact on other projects.

AlexanderSaydakov:
The motivation is to avoid specifying C++ standard all over the place. It should not be necessary.

PR #424, opened 2024-02-21
uint32_t weight in tdigest<float> + cross-language test

AlexanderSaydakov:
generate files for Java


tdigest/include/tdigest.hpp
```diff
@@ -84,18 +85,20 @@ class tdigest {
...
   public:
-    centroid(T value, uint64_t weight): mean_(value), weight_(weight) {}
+    centroid(T value, W weight): mean_(value), weight_(weight) {}
     void add(const centroid& other) {
       weight_ += other.weight_;
       mean_ += (other.mean_ - mean_) * other.weight_ / weight_;
     }
     T get_mean() const { return mean_; }
-    uint64_t get_weight() const { return weight_; }
+    T get_weight() const { return weight_; }
```
jmalkin, line 98: return type W?

tdigest/test/tdigest_test.cpp
```diff
@@ -204,6 +204,27 @@ TEST_CASE("serialize deserialize bytes non empty", "[tdigest]") {
...
+    REQUIRE(((char*)bytes.data())[i] == (char)s.get());
+  }
+
+  s.seekg(0); // rewind
+  auto deserialized_td1 = tdigest<double>::deserialize(s);
+  auto deserialized_td2 = tdigest<double>::deserialize(bytes.data(), bytes.size());
+  REQUIRE(bytes.size() == static_cast<size_t>(s.tellg()));
+
+  REQUIRE(deserialized_td1.get_k() == deserialized_td2.get_k());
+  REQUIRE(deserialized_td1.get_total_weight() == deserialized_td2.get_total_weight());
```
jmalkin, line 224: This never actually checks that the weight is 0 or that `is_empty()` returns true, only that they're equivalent

jmalkin (changes requested):
Can improve one test slightly and I think we need to fix the weight type, unless the intent is to cast it into a floating-point type

jmalkin (approved):
👍

PR #425, opened 2024-03-02
default k=200, special single value serialization format

AlexanderSaydakov:


jmalkin (approved):
sorta weird jumping between java and c++ but at least i recognized some patterns which made it easy to think about consistency.

PR #428, opened 2024-03-12
added get_serialized_size_bytes()

AlexanderSaydakov:


jmalkin (approved):

tdigest/include/tdigest_impl.hpp
```diff
@@ -282,68 +275,76 @@ double tdigest<T, A>::weighted_average(double x1, double w1, double x2, double w
 }
 
 template<typename T, typename A>
-void tdigest<T, A>::serialize(std::ostream& os) const {
-  const_cast<tdigest*>(this)->merge_buffered(); // side effect
+void tdigest<T, A>::serialize(std::ostream& os, bool with_buffer) const {
+  if (!with_buffer) const_cast<tdigest*>(this)->compress(); // side effect
   write(os, is_empty() || is_single_value() ? PREAMBLE_LONGS_EMPTY_OR_SINGLE : PREAMBLE_LONGS_MULTIPLE);
```
jmalkin, line 280: can call the newly added `get_preamble_longs()` here

tdigest/include/tdigest_impl.hpp
```diff
@@ -282,68 +275,76 @@ double tdigest<T, A>::weighted_average(double x1, double w1, double x2, double w
...
+  if (with_buffer) size_bytes += sizeof(T) * buffer_.size(); // count is a part of preamble
+  return size_bytes;
+}
+
+template<typename T, typename A>
+auto tdigest<T, A>::serialize(unsigned header_size_bytes, bool with_buffer) const -> vector_bytes {
+  if (!with_buffer) const_cast<tdigest*>(this)->compress(); // side effect
+  vector_bytes bytes(get_serialized_size_bytes(with_buffer), 0, buffer_.get_allocator());
+  uint8_t* ptr = bytes.data() + header_size_bytes;
+  *ptr++ = get_preamble_longs();
```
jmalkin, line 326: One of the more aggressive warning flags complains about *ptr++ as possibly incrementing a null ptr

tdigest/test/tdigest_test.cpp
```diff
@@ -54,8 +54,9 @@ TEST_CASE("one value", "[tdigest]") {
 
 TEST_CASE("many values", "[tdigest]") {
   const size_t n = 10000;
-  tdigest_double td(100);
+  tdigest_double td;
   for (size_t i = 0; i < n; ++i) td.update(i);
+//  std::cout << td.to_string(true);
```
jmalkin, line 59: not a blocker but some extra comments left in it seems

jmalkin (approved):
A couple very minor comments, not blockers

tdigest/include/tdigest_impl.hpp
AlexanderSaydakov: I don't see how this would happen.
I thought about encapsulating this keeping track of the offset. or, perhaps, using something like std::strstream

tdigest/test/tdigest_test.cpp
AlexanderSaydakov: these are tests, and I uncomment this sometimes to check. not a big deal. we can kill this later

tdigest/include/tdigest_impl.hpp
AlexanderSaydakov: right

jmalkin (approved):

PR #432, opened 2024-05-15
Fix UBSan report while compiling with clang-18

alexey-milovidov:
This is a benign report. But clang-18 started to complain.

See https://github.com/ClickHouse/ClickHouse/issues/63858


jmalkin:
This makes sense. I don't think the existing code is unsafe, but UBSan has trouble determining that and (validly) complains.

PR #438, opened 2024-08-13
Bloom filter

jmalkin:
I think this is a complete filter, including compatibility tests iwht java (that are not yet implemented in java!).

I'm look for feedback on how I handled wrapping. It feels a little too parallel to java in some ways, with `wrap()` vs `writable_wrap()`, which I'm not sure is a good idea. The other issue is taht the former returns a const sketch, but counting the number of bits set may require modifying the filter. It should compute the number if it's in read-only mode so I could potentially add a const version of that method -- or just make it not update the internal state.

Anyway, I'm fairly sure there are missed const values and other changes needed. But feel free to start with big picture design first and then move to details once we're happy with the overall structure.


filters/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+add_library(filters INTERFACE)
+
+add_library(${PROJECT_NAME}::FILTER ALIAS filters)
```
AlexanderSaydakov, line 20: I think we have the same name in caps in all other subdirs. I am not sure what effect this has.

filters/include/bit_array_ops.hpp
```diff
@@ -0,0 +1,180 @@
...
+    // read-only checks handled by set_bit() and clear_bit()
+    if (value) {
+      set_bit(array, index);
+    } else {
+      clear_bit(array, index);
+    }
+  }
+
+  /**
+   * Gets teh value of a bit at the specified index and sets it to true
```
AlexanderSaydakov, line 82: typo

filters/include/bit_array_ops.hpp
```diff
@@ -0,0 +1,180 @@
...
+ */
+namespace bit_array_ops {
+
+  /**
+   * Get the value of a bit at the given index.
+   * @param array the array of bits
+   * @param index the index of the bit to get
+   * @return the value of the bit at the given index.
+   */
+  static inline bool get_bit(uint8_t* array, const uint64_t index) {
```
AlexanderSaydakov, line 45: I think we did not want "const" on copied values

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+   * @param allocator instance of an Allocator
+   * @return a Bloom filter wrapping the provided memory
+   */
+  static bloom_filter_alloc writable_wrap(void* data, size_t length_bytes, const Allocator& allocator = Allocator());
+
+  /**
+   * Copy constructor
+   * @param other filter to be copied
+   */
+  bloom_filter_alloc(const bloom_filter_alloc&);
```
AlexanderSaydakov, line 234: param other is omitted

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+
+  /** Move constructor
+   * @param other filter to be moved
+   */
+  bloom_filter_alloc(bloom_filter_alloc&&) noexcept;
```
AlexanderSaydakov, line 239: param is omitted

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+
+  /**
+   * Move assignment
+   * @param other filter to be moved
+   * @return reference to this filter
+   */
+  bloom_filter_alloc& operator=(bloom_filter_alloc&& other);
+
+  /**
+   * @brief Destroy the bloom filter alloc object
```
AlexanderSaydakov, line 256: I don't think we want to mention alloc here

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+
+  // This is a convenience alias for users
+  // The type returned by the following serialize method
+  using vector_bytes = std::vector<uint8_t, typename std::allocator_traits<A>::template rebind_alloc<uint8_t>>;
+
+  /**
+   * This method serializes the filter as a vector of bytes.
+   * An optional header can be reserved in front of the filter.
+   * It is a blank space of a given size.
+   * This header is used in Datasketches PostgreSQL extension.
```
AlexanderSaydakov, line 268: not sure we want to mention PostgreSQL here

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+  bool is_read_only() const;
+
+  /**
+   * @brief Returns whether the filter owns its underlying memory
+   * @return True if the filter owns its memory, otherwise false
+   */
+  bool is_memory_owned() const;
+
+  /**
+   * @brief Checks if the Bloom Filter has backing memory.
```
AlexanderSaydakov, line 661: not clear to me what does this mean

filters/CMakeLists.txt
jmalkin: I'm not entirely sure either, tbh. I think the official name is `datasketches::FILTER` and we alias it to `filter`?

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+   */
+  static bloom_filter_alloc deserialize(const void* bytes, size_t length_bytes, const Allocator& allocator = Allocator());
+
+  /**
+   * This method deserializes a Bloom filter from a given stream.
+   * @param is input stream
+   * @param allocator instance of an Allocator
+   * @return an instance of a Bloom filter
+   */
+  static bloom_filter_alloc deserialize(std::istream& is, const A& allocator = Allocator());
```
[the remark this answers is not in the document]
jmalkin: Right. I mean to use `Allocator` in this file and `A` in the _impl files

filters/include/bloom_filter.hpp
jmalkin: We did on all other sketches?

filters/include/bloom_filter.hpp
jmalkin: would `is_wrapped()` work better?

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+  bool internal_query_and_update(const uint64_t h0, const uint64_t h1);
+  bool internal_query(const uint64_t h0, const uint64_t h1) const;
+
+  void update_num_bits_set(const uint64_t num_bits_set);
+
+  A allocator_;
+  uint64_t seed_;
+  uint16_t num_hashes_;
+  bool is_dirty_;
+  bool is_owned_; // if true, data is not owned by filter AND data_ holdes the entire filter not just the bit array
```
AlexanderSaydakov, line 742: "holdes"

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,757 @@
...
+  bool is_read_only_; // if true, filter is read-only
+  uint64_t capacity_bits_;
+  uint64_t num_bits_set_;
+  uint8_t* bit_array_;  // data backing bit_array_, regardless of ownership
+  uint8_t* memory_; // if wrapped, pointer to the start of the filter, otheriwse nullptr
+
+  friend class bloom_filter_builder_alloc<A>;
```
AlexanderSaydakov, line 749: don't we want to declare the builder in this class as we do in Theta sketch?

filters/include/bloom_filter_builder_impl.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+namespace datasketches {
+
+template<typename A>
+uint64_t bloom_filter_builder_alloc<A>::generate_random_seed() {
+  union {
+    uint64_t long_value;
+    double double_value;
+  } ldu;
+  ldu.double_value = random_utils::next_double(random_utils::rand);
```
AlexanderSaydakov, line 38: this reinterpretation of double [0, 1] as uint64_t seems a bit too tricky. I am not sure it actually produces a number uniformly distributed in the 64-bit range.
perhaps, std::uniform_int_distribution would be more straightforward?

filters/include/bloom_filter_impl.hpp
```diff
@@ -0,0 +1,907 @@
...
+bloom_filter_alloc<A>::bloom_filter_alloc(bloom_filter_alloc&& other) noexcept :
+  allocator_(std::move(other.allocator_)),
+  seed_(other.seed_),
+  num_hashes_(other.num_hashes_),
+  is_dirty_(other.is_dirty_),
+  is_owned_(other.is_owned_),
+  is_read_only_(other.is_read_only_),
+  capacity_bits_(other.capacity_bits_),
+  num_bits_set_(other.num_bits_set_),
+  bit_array_(std::move(other.bit_array_)),
```
AlexanderSaydakov, line 185: move doesn't make much sense for a pointer. this is just a copy

filters/include/bloom_filter_builder_impl.hpp
jmalkin: Ooh.. Thanks. It definitely does not produce a valid uniform distribution over 64-bit values.

filters/test/bloom_filter_deserialize_from_java_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+static std::string testBinaryInputPath = std::string(TEST_BINARY_INPUT_PATH) + "../../java/";
+
+TEST_CASE("bloom_filter", "[serde_compat]") {
+  const uint64_t n_arr[] = {0, 10000, 2000000, 30000000};
+  const double h_arr[] = {3, 5};
+  for (const uint64_t n: n_arr) {
+    for (const uint16_t num_hashes: h_arr) {
+      std::ifstream is;
+      is.exceptions(std::ios::failbit | std::ios::badbit);
+      is.open(testBinaryInputPath + "bf_n" + std::to_string(n) + "_h" + std::to_string(num_hashes) + "_cpp.sk", std::ios::binary);
```
AlexanderSaydakov, line 38: seems like should be "_java.sk"

filters/CMakeLists.txt
AlexanderSaydakov: should this be the same as the subdir?

filters/include/bloom_filter.hpp
AlexanderSaydakov: I think we did. but I am not sure it was good, especially for a sketch that is not supported in PSQL
perhaps we could soften the statement "is used". something like "such headers might be needed in some systems like PSQL"

filters/include/bloom_filter.hpp
AlexanderSaydakov: not sure. this is the first mention of backing, and no explanation. I am still not sure what that backing is having glanced through the whole PR

filters/test/bloom_filter_deserialize_from_java_test.cpp
jmalkin: d'oh... yes, i was initially testing on its own output and forgot to ever fix that

filters/include/bit_array_ops.hpp
jmalkin: The main advantage is that the compiler ensures you're not going to inadvertently try writing to a value that should never change. But that's fine. I'll remove them in many places.

filters/include/bloom_filter.hpp
jmalkin: How is "Some integrations such as PostgreSQL may need this header space."

filters/include/bloom_filter.hpp
jmalkin: ok, changed to `is_wrapped()` and `get_wrapped_memory()` which should be more clear

filters/include/bloom_filter_impl.hpp
jmalkin: yeah, i was trying to use copilot and while it helps with some boilerplate it inserts stuff like this, too, and for whatever reason i missed removing it.  thinking i need to turn that off overall.

filters/CMakeLists.txt
jmalkin: missing an S? probably? but only if we also put quotient and infinifilter in there.  otherwise this should probably live elsewhere.

AlexanderSaydakov: I think we will put quotient filter there. and eventually infinifilter

filters/include/bit_array_ops.hpp
AlexanderSaydakov: yes, perhaps we should do that, but that is not our current style

filters/include/bloom_filter.hpp
AlexanderSaydakov: looks fine

filters/CMakeLists.txt
jmalkin: still not sure it's actually needed or even best practice, but added the S

filters/include/bloom_filter.hpp
jmalkin: Agnostic on this, so made it parallel to theta for consistency

jmalkin:
I think I have addressed all the comments now

filters/include/bloom_filter.hpp
```diff
@@ -0,0 +1,756 @@
...
+#include <memory>
+#include <vector>
+
+#include "common_defs.hpp"
+
+namespace datasketches {
+
+// forward declarations
+template<typename A> class bloom_filter_alloc;
+template<typename A> class bloom_filter_builder_alloc;
```
AlexanderSaydakov, line 34: do we still need this one?

AlexanderSaydakov (approved):

filters/include/bloom_filter.hpp
jmalkin: We do not. Removed, and will wait for tests to pass before merging,

PR #442, opened 2024-10-15
explicit java setup

AlexanderSaydakov:


jmalkin (approved):
we're definitely going to forget to update this when we move the java repo to 17...

AlexanderSaydakov:
This is to fix the failing serde compatibility action, that seems to have moved to Java 17 by default (Ubuntu 24.04 LTS)

PR #447, opened 2025-01-09
Fix bit packing

AlexanderSaydakov:


theta/test/bit_packing_test.cpp
```diff
@@ -29,51 +29,53 @@ namespace datasketches {
...
-    std::vector<uint8_t> bytes(n * sizeof(uint64_t), 0);
-    uint8_t offset = 0;
-    uint8_t* ptr = bytes.data();
-    for (int i = 0; i < n; ++i) {
-      offset = pack_bits(input[i], bits, ptr, offset);
-    }
+  uint64_t value = 0xaa55aa55aa55aa55ULL; // arbitrary starting value
+  for (int n = 0; n < 100; ++n) {
+    for (uint8_t bits = 1; bits <= 63; ++bits) {
+      int n = 8;
```
jmalkin, line 35: this conflicts with the for loop 2 lines above

jmalkin (commented):
I think one of the tests will not quite do what we want

jmalkin (approved):

PR #458, opened 2025-06-07
Provide get_centroids implementation for tdigest

MaheshGPai:
This is required for providing a mechanism to share tdigest across other languages. Centroids can be used to rebuild the tdigest.


jmalkin:
Please see https://apache.github.io/datasketches-java/6.2.0/org/apache/datasketches/tdigest/TDigestDouble.html for a binary-compatible implementation

MaheshGPai:
@jmalkin Thanks for the link. I understand that serialized bytes can be used to build the tdigest in jaav. But there are cases where [different Tdigest library](https://github.com/tdunning/t-digest) is used. For such cases providing the list of centroids are useful.
Is this something that can be exposed?

jmalkin:
If you're insistent on trying to move things in ways that risk breaking data fidelity this should be done as a read-only iterator on the class, not by return the internal private object

AlexanderSaydakov:
By the way, why the other library "is required to be used"?

MaheshGPai:
> By the way, why the other library "is required to be used"?

Some of the consumers use this library extensively in the downstream pipelines and something that cannot be immediately updated to replace with the datasketch library.

tdigest/include/tdigest_impl.hpp
```diff
@@ -627,6 +627,65 @@ void tdigest<T, A>::check_split_points(const T* values, uint32_t size) {
...
+}
+
+template <typename T, typename A>
+  typename tdigest<T, A>::const_iterator tdigest<T, A>::end() const {
+  return tdigest::const_iterator(*this, true);
+}
+
+template<typename T, typename A>
+tdigest<T, A>::const_iterator::const_iterator(const tdigest& tdigest_, const bool is_end):
+  centroids_()
```
AlexanderSaydakov, line 642: this seems to break the AllocatorAwareContainer requirement

tdigest/include/tdigest_impl.hpp
```diff
@@ -627,6 +627,65 @@ void tdigest<T, A>::check_split_points(const T* values, uint32_t size) {
...
+{
+  // Create a copy of the tdigest to generate the centroids after processing the buffered values
+  tdigest tmp(tdigest_);
```
AlexanderSaydakov, line 645: this should also support allocator

tdigest/test/tdigest_test.cpp
```diff
@@ -453,4 +453,18 @@ TEST_CASE("deserialize from reference implementation bytes float", "[tdigest]")
...
+  for (int i = 0; i < 10; i++) {
+    td.update(i);
+  }
+
+  auto centroid_count = 0;
+  for (const auto &centroid: td) {
+    centroid_count++;
+  }
+  // Ensure that centroids are retrieved for a case where there is buffered values
+  REQUIRE(centroid_count == 10);
```
AlexanderSaydakov, line 467: perhaps we could do better than just counting, but I am not sure whether that would be assuming too much about implementation

tdigest/include/tdigest_impl.hpp
MaheshGPai: Fixed

tdigest/test/tdigest_test.cpp
MaheshGPai: I have added validation to check the weights as well. Cant think of anything better. Please do let me know if you have anything specific in mind.

tdigest/include/tdigest_impl.hpp
```diff
@@ -597,6 +597,24 @@ bool tdigest<T, A>::is_single_value() const {
   return get_total_weight() == 1;
 }
 
+template<typename T, typename A>
```
AlexanderSaydakov, line 600: why? what does this do differently compared to the implicit one?

MaheshGPai: Maybe my understanding is wrong. I assumed this change is required to pass on the allocators to the members.

tdigest/include/tdigest_impl.hpp
AlexanderSaydakov: I think I was wrong here. Copy constructor should do the right thing.

tdigest/include/tdigest_impl.hpp
```diff
@@ -638,11 +656,11 @@ template <typename T, typename A>
...
 
 template<typename T, typename A>
-tdigest<T, A>::const_iterator::const_iterator(const tdigest& tdigest_, const bool is_end):
-  centroids_()
+tdigest<T, A>::const_iterator::const_iterator(const tdigest<T, A>& tdigest_, const bool is_end):
+  centroids_(tdigest_.get_allocator())
 {
   // Create a copy of the tdigest to generate the centroids after processing the buffered values
-  tdigest tmp(tdigest_);
+  tdigest<T, A> tmp(tdigest_);
```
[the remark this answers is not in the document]
AlexanderSaydakov: Could you also remove the unnecessary parameters <T, A> in two places above? This is nit picking, but let's be consistent while we still remember.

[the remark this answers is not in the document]
MaheshGPai: Done

AlexanderSaydakov:
I am still not sure whether this is the best way to provide compatible serialization. Perhaps a better way would be to have serialize_compat() methods or something like that.
Having said that, I don't mind this const iterator over the array of centroids. Arguably, it is similar to other quantile sketches, and should provide a set of "best" points representing the distribution seen by t-digest (best that t-digest can do, presumably).

tdigest/include/tdigest.hpp
```diff
@@ -297,6 +310,28 @@ class tdigest {
...
+template<typename T, typename A>
+class tdigest<T, A>::const_iterator {
+public:
+  using iterator_category = std::input_iterator_tag;
+  using value_type = std::pair<const T&, const W>;
+  using difference_type = void;
+  using pointer = const return_value_holder<value_type>;
+  using reference = const value_type;
+
+  const_iterator(const tdigest<T, A> &tdigest_, bool is_end);
```
AlexanderSaydakov, line 322: I think this constructor should be private

AlexanderSaydakov: and <T, A> seems unnecessary again.
and our style is to attach & to the type, not to the name.
should be the last nit to pick :)

MaheshGPai: Done

AlexanderSaydakov (approved):

PR #471, opened 2026-01-07
fix: rejecting inf as value

proost:
Currently I'm writing code for go t-Digest. 

While I refer C++ implementation, i found that no guard for +inf, -inf.

Additional Question: There is no validation logic in the t-Digest about overflow or underflow from calculation. Is intended behavior?

### DIFF
--- a/tdigest/include/tdigest.hpp
+++ b/tdigest/include/tdigest.hpp
@@ -154,6 +154,7 @@ public:
    * Compute approximate normalized rank of the given value.
    *
    * <p>If the sketch is empty this throws std::runtime_error.
+   * <p>NaN value throw std::invalid_argument.
    *
    * @param value to be ranked
    * @return normalized rank (from 0 to 1 inclusive)


tdunning:
@proost Let's talk. Feel free to ping me directly on my apache email address. I would love to work with you on a Go implementation and there are important simplifications in Go relative to the Java implementation because you can easily have an array of structs.

Regarding your question, a guard for ±inf is more subtle than it looks. If you insert a single infinite value, it will have a centroid with one data point in it. That's fine. If you insert another on the same side, you will start to get centroids with infinite value which is actually numerically correct. You will also have effects on the interpolation code, but that is also likely fairly reasonable.

How do you think an empirical distribution with infinite values should be handled?

What do you think the mean of such a distribution should be?

proost:
@AlexanderSaydakov 

There is one thing I wanted to clarify regarding this PR.

In the current implementation, ±∞ values can flow into the t-digest and be treated as data points. With this PR, ±∞ values passed to `update()` are silently ignored, and ±∞ passed to query methods such as `get_rank()` result in an exception. That means this change alters the behavior for users who might already be passing infinity values (intentionally or not), which effectively makes it a breaking change.

If supporting ±∞ as valid data was intentional in the original design, then I agree that changing this behavior is not appropriate. However, if it wasn’t an intentional design choice and infinity handling was simply unspecified, then I think explicitly rejecting/ignoring infinities and documenting the behavior would improve robustness and make the policy clearer for users.

I wanted to check your view on this before moving forward, since it affects compatibility and library semantics.

tisonkun:
> Arithmetic with infinity in IEEE floating point *can* produce NaN, but only when opposing values are involved. Sorting works correctly because infinities compare with normal numbers. Since the centroids are always sorted, merging will only ever combine infinities of like sign. This means that infinity will creep slowly like a contagion into the centroids in the digest. If there are lots of infinite values in a sample, we may get a slight over-estimate of quantiles because an entire centroid will go to infinity if a single infinite value is introduced, but no single centroid is that large in a reasonably configured digest.

In general inputs, we shall never see NaN, Inf, -Inf, or at least not too much.

We're here to discuss edge cases. So the trade-off is that, assuming the sketches may have +-inf but they never combine to produce NaN, or just filter all +-inf.

If we allow +-inf and assume they will never combine due to intermediate finite numbers, the question is how we program against the opposing values? Shall we mark the sketch in a broken state, and all operations then fail? Or crash the program? (Since NaN cannot compare with other values; or we can define a total order in some way.)

If we declare that this won't happen, then it means that is an undefined behavior.

tdunning:
Your key statement is that this never (practically speaking) happens.

That means that filtering is not a breaking change (practically speaking),
nor does it produce different results for any reasonable input.

Filtering out invalid inputs makes it easier to reason about the code and
so it can actually improve things. Pretending that we have covered all the
bases in the presence of infinite values is probably not a realistic thing
to do.

You have convinced me.

proost:
I reopen this PR.

I refer rust code, so I added missing validation which is deserialization. Thanks to @tisonkun .

I also add comments that what is expected and valid input too.

tdigest/include/tdigest_impl.hpp
```diff
@@ -94,6 +109,7 @@ template<typename T, typename A>
 double tdigest<T, A>::get_rank(T value) const {
   if (is_empty()) throw std::runtime_error("operation is undefined for an empty sketch");
   if (std::isnan(value)) throw std::invalid_argument("operation is undefined for NaN");
+  if (std::isinf(value)) throw std::invalid_argument("operation is undefined for infinity");
```
tisonkun:
```suggestion
```

`get_rank` can accept `+-inf` as my previous comment noted. It doesn't change the internal state.

Also update the method document that `NaN` is rejected but infinite values are not.

proost: [662aef3](https://github.com/apache/datasketches-cpp/pull/471/commits/662aef37c3912b4b2c6cbf3cc6ab0dbf5d40a0df)

Oh, You are right! Thank you!

Issue #475, opened 2026-01-25
[feature] Array of Strings tuple sketch


proost:
# Summary

Adding Array of Strings Tuple Sketch.

# Design

- I reuse and extend the existing tuple sketch interfaces as much as possible.
- I use `array<std::string>` as container for string collection, following the same approach as the ArrayOfDoubles tuple sketch.
- Serialization and deserialization follow the existing tuple sketch patterns using `default_array_of_strings_serde` .
    - One difference from the Java implementation is UTF-8 validation. In the C++ ArrayOfStringsTupleSketch, I added validation to ensure each std::string contains valid UTF-8 encoded text. This difference comes from the fact that Java String is always valid Unicode, while C++ std::string may contain arbitrary bytes.
    - For validation, I use utfcpp (https://github.com/nemtrif/utfcpp). simdutf (https://github.com/simdutf/simdutf) is widely used and C++11 compatible, but it may trigger warnings because it does not strictly follow ISO C++ in all configurations. Since some users build with strict compiler flags and UTF-8 validation is not on the hot path (it runs only during serialization/deserialization), I chose utfcpp even though it may be slightly slower.

```cpp
namespace datasketches {

// default update policy for an array of strings
template<typename Allocator = std::allocator<std::string>>
class default_array_of_strings_update_policy {
public:
  using array_of_strings = array<std::string, Allocator>;

  explicit default_array_of_strings_update_policy(const Allocator& allocator = Allocator());

  array_of_strings create() const;

  void update(array_of_strings& array, const array_of_strings& input) const;

  void update(array_of_strings& array, const array_of_strings* input) const;

private:
  Allocator allocator_;
};

// serializer/deserializer for an array of strings
// Requirements: all strings must be valid UTF-8 and array size must be <= 127.
template<typename Allocator = std::allocator<std::string>>
struct default_array_of_strings_serde {
  using array_of_strings = array<std::string, Allocator>;
  using summary_allocator = typename std::allocator_traits<Allocator>::template rebind_alloc<array_of_strings>;

  explicit default_array_of_strings_serde(const Allocator& allocator = Allocator());

  void serialize(std::ostream& os, const array_of_strings* items, unsigned num) const;
  void deserialize(std::istream& is, array_of_strings* items, unsigned num) const;
  size_t serialize(void* ptr, size_t capacity, const array_of_strings* items, unsigned num) const;
  size_t deserialize(const void* ptr, size_t capacity, array_of_strings* items, unsigned num) const;
[...]

PR #476, opened 2026-01-27
feat: AoS tuple sketch

proost:
issue: https://github.com/apache/datasketches-cpp/issues/475

I change original design. I will follow same design with array of doubles sketch, that allow allocator for container not the element of container. I think using standard allocator for string is fine, for consistency of code following same design is good And even if we use custom allocator, we can't use completely for string allocation because of serialization and deserialization. 

even though above reasons, if want to change to orignal design, it is acceptable.


proost:
@tisonkun 
If you have time, can you review this PR?

proost:
@leerho 
I reflected from discussion.

I will send other PRs for serde and other sketches.

PR #479, opened 2026-01-31
fix: Add the missing brackets and keep the class member name consistency

SYaoJun:
use clang-tidy fix the warning

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,7 +27,7 @@ Checks: |
 
 CheckOptions:
   - key:             google-readability-braces-around-statements.ShortStatementLines
-    value:           '1'
+    value:           '0'
   - key:             google-readability-function-size.StatementThreshold
     value:           '800'
   - key:             google-readability-namespace-comments.ShortNamespaceLines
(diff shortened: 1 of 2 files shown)


SYaoJun:
> I am not sure I support changing short one-line if statements like this from
> 
> if (condition) return 0;
> 
> to
> 
> if (condition) { return 0; }
> 
> We did not establish any style requirements, and we use one-liners throughout the C++ library. So why should we change the approach in this particular class?
> 
> There is no harm, of course, but this style is a bit more verbose. I would not mind switching to this style if the community wants.

I appreciate your perspective. While this individual style change might seem minor, I believe we should think bigger: implementing a `.clang-tidy` configuration would provide long-term benefits for code consistency and maintainability.

I'd like to propose we adopt a `.clang-tidy` from another Apache project as a reference, then refactor this code to comply with it. As the codebase grows, having explicit linting rules becomes essential. Without them, we risk inconsistency and harder maintenance down the road.

Does this sound reasonable to you? We could start by researching configurations from similar Apache projects.

AlexanderSaydakov:
While I personally prefer brief one-line style, I don't mind adopting a stricter approach with mandatory curly brackets if we consistently use it throughout the whole code in C++.

leerho:
I agree with the proposed .clang-tidy configuration.

SYaoJun:
I have updated the code to support one line statement.

PR #481, opened 2026-02-02
test: add missing KLL long sketch compatibility cases

proost:
I found that datasketches-java generate long sketches for compatibility test but C++ doesn't have anything.

After merged, I also send a PR for long type case.


leerho (approved):
LGTM

PR #484, opened 2026-02-13
fix: wrong k condition check

proost:
we should throw exception when k is negative.


proost:
@jmalkin 
I'm sorry, i missed type in signature during comparing to java code.

jmalkin:
It happens to everyone at times :)

Issue #499, opened 2026-04-10
Compilation failure with GCC 15


mcgrawia:
**Description:**

datasketches-cpp 5.1.0 fails to compile under GCC 15 due to a [documented libstdc++ change](https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes) that removed transitive includes of `<cstdint>`. Several internal libstdc++ headers (`bits/align.h`, `bits/atomic_base.h`, `bits/atomic_wait.h`) no longer include `<stdint.h>`, so fixed-width integer types like `uint8_t`, `uint16_t`, `uint32_t`, and `uint64_t` are no longer available unless explicitly included.

`fi/include/reverse_purge_hash_map.hpp` uses these types throughout (template defaults, member variables, function parameters, constants) but never directly includes `<cstdint>`. Under GCC 14 and earlier these types were available transitively; under GCC 15 they are not, causing a cascade of compilation errors.

This is the same class of breakage that affected [Node.js](https://github.com/nodejs/node/pull/56740), [Godot](https://github.com/godotengine/godot/pull/102022), [PowerDNS](https://github.com/PowerDNS/pdns/pull/15520), and [many other projects](https://gcc.gnu.org/gcc-15/changes.html) after GCC 15 shipped.

**First error:**

```
reverse_purge_hash_map.hpp:38:16: error: 'uint64_t' does not name a type
   38 |   typename V = uint64_t,
      |                ^~~~~~~~
reverse_purge_hash_map.hpp:25:1: note: 'uint64_t' is defined in header '<cstdint>';
  this is probably fixable by adding '#include <cstdint>'
```

**Reproducer:**

Build datasketches-postgresql 1.7.0 (which uses datasketches-cpp 5.1.0) on any platform with GCC 15:

```dockerfile
FROM timescale/timescaledb:2.26.1-pg16  # Alpine 3.23, GCC 15.2.0
# standard datasketches-postgresql build steps...
# Fails at: make
```

**Suggested fix:**

Add `#include <cstdint>` to `fi/include/reverse_purge_hash_map.hpp` (after the existing `#include <iterator>` on line 24). There may be other headers in the project with the same issue.

**Environment:**
- GCC 15.2.0 / libstdc++ (Alpine 3.23)
- datasketches-cpp 5.1.0
- datasketches-postgresql 1.7.0
- PostgreSQL 16

PR #500, opened 2026-04-11
fix: missing header with gcc 15

proost:
issue: https://github.com/apache/datasketches-cpp/issues/499

I don't know exactly coverage what we covered. **setting coverage is needed for ds-cpp project**

Following the current CI machine(ubuntu 24.04), I tried to test all current supported gcc and new gcc 15.

And also we tried to cover from C++11, so add them too.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -6,7 +6,7 @@ env:
   BUILD_TYPE: Release
 
 jobs:
-  build:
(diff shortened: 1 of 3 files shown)


leerho:
Can you confirm that this will still compile with C++ 11?

leerho (approved):
This passes the CI checks, which are based on C++11, so this should be good to go.

PR #504, opened 2026-06-18
Update .gitattributes for new file types

tisonkun:
See https://github.com/apache/datasketches-rust/pull/92#issuecomment-4743982760.

I don't know if it works but it should at least do no harm.


proost (approved):
Thank you!

PR #505, opened 2026-06-24
Fix count-min sketches test snapshots should use binary mode

tisonkun:
... otherwise `\n` would be `\r\n` on Windows and cause https://github.com/apache/datasketches-rust/pull/92 to fail.

### DIFF
--- a/count/test/count_min_test.cpp
+++ b/count/test/count_min_test.cpp
@@ -215,7 +215,7 @@ TEST_CASE("CountMin sketch: serialize-deserialize empty", "[cm_sketch]") {
     for (auto di: d) {
       REQUIRE(di == 0);
     }
-    std::ofstream os("count_min-empty.bin", std::ios::binary);
+    std::ofstream os("count_min_empty_cpp.sk", std::ios::binary);
     c.serialize(os);
 }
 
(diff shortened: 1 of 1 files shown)


tisonkun:
Seems other snapshot generations already use binary mode.

PsiACE:
since it looks like all the other files end with ".sk" , only count min with `.bin`

tisonkun:
> since it looks like all the other files end with ".sk" , only count min with `.bin`

Good point. As other impls may also depend on the name, I'd sync with @leerho @proost @freakyzoidberg first to make the unify if any.

proost (approved):
Thank you! I agree with changing file extension to ".sk".

tisonkun:
> Thank you! I agree with changing file extension to ".sk".

I agree with this. I can update it in this PR or do it in a follow-up.

One possible impact is that datasketches-java and datasketches-go, IIRC, test against these snapshot files. So they may be affected by the name changes as well. But for sure we can update there cascadingly, just need to inform all the related devs.

tisonkun:
Pushed 7c993ed201dced4458ee72954642aabfa80708c8 for the rename.

I'm going to merge this patch later today or tomorrow and fix any impact downstream.

PR #506, opened 2026-07-02
Remove allocation in `count_min_sketch::get_hashes`

cv4g:
Hi,

the countmin sketch is in use at [ClickHouse](https://github.com/ClickHouse/ClickHouse) and we noticed that there is a vector allocated on the hot path in `count_min_sketch::get_hashes`. To avoid it one could either inline it's body into the call sites or modify it to take a lambda that is invoked on each hash. Would you be interested in merging one of both approaches?

In a microbenchmark the change seems to yield an improvement for update between factor 1.1 and 1.4 and for estimate between factor 1.7 and 3.5:

```
   Group                    Time       CPU
  ─────────────────────  ────────  ────────
   UpdateUInt64           1.440x    1.440x
  ─────────────────────  ────────  ────────
   UpdateStringBytes      1.118x    1.119x
  ─────────────────────  ────────  ────────
   EstimateUInt64         3.548x    3.549x
  ─────────────────────  ────────  ────────
   EstimateStringBytes    1.764x    1.764x
```

The numbers were produced by a `google-benchmark` setup in #507.

Before:
```
Benchmark                                     Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------------------
BM_CountMinUpdateUInt64/1024              19585 ns        19606 ns        35834 bytes_per_second=398.467M/s items_per_second=52.2279M/s
BM_CountMinUpdateUInt64/4096              78486 ns        78515 ns         9017 bytes_per_second=398.011M/s items_per_second=52.1681M/s
BM_CountMinUpdateUInt64/32768            629480 ns       629521 ns         1103 bytes_per_second=397.127M/s items_per_second=52.0523M/s
BM_CountMinUpdateUInt64/65536           1252762 ns      1252815 ns          554 bytes_per_second=399.101M/s items_per_second=52.311M/s
BM_CountMinUpdateStringBytes/1024         34527 ns        34571 ns        20270 bytes_per_second=722.64M/s items_per_second=29.6201M/s
BM_CountMinUpdateStringBytes/4096        137133 ns       137181 ns         5092 bytes_per_second=739.661M/s items_per_second=29.8584M/s
BM_CountMinUpdateStringBytes/32768      1120784 ns      1120869 ns          632 bytes_per_second=749.194M/s items_per_second=29.2345M/s
BM_CountMinUpdateStringBytes/65536      2250362 ns      2250433 ns          310 bytes_per_second=759.885M/s items_per_second=29.1215M/s
[...]

PR #507, opened 2026-07-02
Add micro-benchmark for `count_min_sketch`

cv4g:
Hi,

for #506 I used a micro-benchmark to determine the speed up of a change in `count_min_sketch`. Would you deem a micro-benchmarking library like `google-benchmark` useful in `datasketches-cpp`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,8 @@ if (BUILD_TESTS)
   enable_testing()
 endif()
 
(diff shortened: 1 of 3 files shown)


proost (commented):
can you add `benchmark::ClobberMemory()` after nested loop in the update case?

proost (approved):
I verified after benchmark code is changed.

In my case(Apple silicon, clang 17), much faster than you suggested.

Thank you!

PR #508, opened 2026-07-07
test: cover req sketch exact mode from serialized bytes

proost:
See this PR: https://github.com/apache/datasketches-java/pull/734

C++ is OK. But to assure compatibility, add test code.


leerho (approved):
LGTM

PR #510, opened 2026-07-20
Remove stale Python CMake hook

PointKernel:
This PR removes the stale `WITH_PYTHON` block from the root CMake configuration. The Python bindings were removed from this repository in #393 after moving to their own project, but the conditional remained and tries to add the nonexistent `python` subdirectory when enabled.


leerho (approved):
LGTM

PR #511, opened 2026-08-10
Bounds-check per-entry key read in compact_tuple_sketch::deserialize

SavicStefan:
The entry-reading loop reads each entry as an 8-byte key followed by a Summary-width summary. The summary read is passed the remaining capacity and bounds-checks itself, but the key read is an unchecked fixed-size copy_from_mem that relies solely on the pre-loop keys-only reservation, which assumes every entry's summary occupies the width Summary serializes to.

When that assumption does not hold (a truncated buffer, or a buffer whose serialized summary width differs from the Summary it is deserialized as) the read cursor advances past the data. num_entries is read from the preamble and is unaffected, so the loop runs the full count and a later key read walks off the end of the buffer: silent on a normal build, a heap-buffer-overflow under AddressSanitizer.

Add an ensure_minimum_memory check before the key read, mirroring the remaining-capacity check the summary read already performs and the up-front size validation the compact theta parser does. A malformed buffer now throws std::out_of_range instead of reading out of bounds.

Adds a tuple_sketch_test case that deserializes a float-summary sketch as a double-summary sketch and asserts std::out_of_range.

### DIFF
--- a/tuple/test/tuple_sketch_test.cpp
+++ b/tuple/test/tuple_sketch_test.cpp
@@ -371,4 +371,19 @@ TEST_CASE("filter", "[tuple_sketch]") {
   }
 }
 
+TEST_CASE("tuple sketch: deserialize with mismatched summary width", "[tuple_sketch]") {
+  // A compact sketch serialized with a narrower summary (float, 4 bytes) and then
+  // deserialized as a wider summary (double, 8 bytes). The per-entry stride the reader
+  // assumes (8-byte key + 8-byte summary) is larger than the entries actually occupy
+  // (8-byte key + 4-byte summary), so the read cursor advances past the end of the buffer.
+  // num_entries is read from the preamble and is unaffected, so the entry loop still runs
+  // the full count and the per-entry key read walks off the end. This must throw rather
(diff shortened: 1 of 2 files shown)


tuple/test/tuple_sketch_test.cpp
```diff
@@ -371,4 +371,19 @@ TEST_CASE("filter", "[tuple_sketch]") {
   }
 }
 
+TEST_CASE("tuple sketch: deserialize with mismatched summary width", "[tuple_sketch]") {
```
proost: Could you rename this test to describe the behavior being protected?
For example:"tuple sketch: deserialize bounds-checks each entry key"

“Mismatched summary width” could imply that the deserializer reliably detects incompatible summary types, which the wire format cannot guarantee.

proost (commented):
The implementation change looks appropriate to me. But I suggest revising the test case to focus on truncated/malformed-buffer safety.

tuple/test/tuple_sketch_test.cpp
SavicStefan: changed it

proost (approved):
Thank you!

PR #512, opened 2026-08-13
Fix HLL union estimate/serialization regression from lazy KxQ rebuild

SavicStefan:
Since the lazy KxQ/curMin rebuild was introduced ("skip updating kxq in HLL merge"), Hll8Array::mergeHll only sets rebuild_kxq_curmin_ and defers the recompute. Several consumers read that deferred state without honoring the flag, so unioning HLL sketches with different lgConfigK could:

- discard accumulated data: isEmpty() reads curMin_/numAtCurMin_, which stay at empty-sketch defaults after a downsampling merge, so a populated gadget looks empty and the next union update overwrites it;
- corrupt the estimate: internalCouponUpdate does incremental KxQ/HIP updates against the stale base;
- serialize non-deterministically: copyAs() rebuilds via register replay (convertToHll8) while estimate/bounds rebuild via the direct sum, so equivalent merges could serialize to different bytes.

The effect was estimates collapsing to ~one input's cardinality (merge-order dependent) and merge-order-dependent serialization; 4.0.1 had neither.

Fix, keeping the lazy-merge optimization:
- isEmpty(): a pending rebuild means the array is non-empty.
- internalCouponUpdate(): rebuild before an incremental update reads KxQ, only when the coupon changes a register (duplicate coupons stay lazy).
- copyAs(): make every same-type result use the direct-sum rebuild so all paths agree, keeping serialization merge-order independent.

Adds a regression tests covering estimate order-independence, is_empty() after a downsampling merge, scalar-after-merge accumulation, and merge-order-independent serialization.

### DIFF
--- a/hll/include/Hll8Array-internal.hpp
+++ b/hll/include/Hll8Array-internal.hpp
@@ -98,6 +98,10 @@ void Hll8Array<A>::internalCouponUpdate(uint32_t coupon) {
 
   const uint8_t curVal = this->hllByteArr_[slotNo];
   if (newVal > curVal) {
+    // A prior HLL merge may have deferred rebuilding the estimator state. Rebuild only when
+    // this coupon changes a register, immediately before the incremental update needs KxQ and
+    // numAtCurMin_. Duplicate coupons therefore preserve the lazy-merge optimization.
+    if (this->rebuild_kxq_curmin_) this->check_rebuild_kxq_cur_min();
     this->hllByteArr_[slotNo] = newVal;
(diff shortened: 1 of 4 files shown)


proost (commented):
I agree on overall, but how about changing `get_result` in union? Like [java one](https://github.com/apache/datasketches-java/blob/master/src/main/java/org/apache/datasketches/hll/Union.java#L791), rebuild before get_result.

And rebuilding in `copy_as` changes logic semantically.

proost (approved):
Thank you!

PR #513, opened 2026-08-14
Add get_live_bytes() to report the current heap footprint of theta, tuple and kll sketches

stojkomilos:
## What

Adds a public `get_live_bytes()` accessor that reports, in O(1), the exact heap a sketch's
mutable internal state is holding **right now**. The sketches already expose
`get_serialized_size_bytes()`, but that is an upper bound on the *serialized* form; there is
currently no way to ask how much memory a live, mutable sketch is actually retaining. This is
useful for memory accounting when many mutable sketches are held in memory at once (e.g. one
sketch per group/window in a streaming aggregation).

## Where

- **theta / tuple** — the accessor lives on `theta_update_sketch_base`, the hash table shared by
  theta and tuple update sketches, unions and intersections. It returns
  `(1 << lg_cur_size_) * sizeof(Entry)`, or `0` when the table is unallocated (`lg_cur_size_ == 0`
  leaves `entries_ == nullptr`). `update_theta_sketch`, `theta_union`, `theta_intersection`, the
  tuple update sketch, `tuple_union` and `tuple_intersection` forward to it. Tuple's union and
  intersection reuse the theta base as their `State`, so no separate implementation is needed
  there.
- **kll** — not a hash table, so `kll_sketch::get_live_bytes()` sums the `items_` buffer
  (`items_size_` elements) and the `levels_` vector, mirroring the sketch's `allocate()` sites.

Unlike `get_serialized_size_bytes()`, this touches no data and does no iteration; it just reports
the current allocation size.

## Tests

Adds Catch2 cases in `theta_sketch_test`, `theta_union_test`, `theta_intersection_test`,
`tuple_sketch_test` and `kll_sketch_test` that:
- check the footprint is a power-of-two count of entries where applicable,
- confirm an unallocated intersection reports `0`,
- confirm the footprint only grows (never shrinks) as entries are inserted and ends up larger than
  the initial allocation.

All theta / tuple / kll suites pass locally.

## Notes

Scope is theta, tuple and kll. `hll` and `cpc` could get an equivalent accessor in a follow-up if
maintainers want the whole family consistent.

### DIFF
--- a/kll/include/kll_sketch.hpp
+++ b/kll/include/kll_sketch.hpp
@@ -411,6 +411,18 @@ class kll_sketch {
     size_t get_serialized_size_bytes(const SerDe& sd = SerDe()) const;
 
     /**
+     * Returns the number of bytes currently allocated on the heap by this sketch: the items_ buffer
+     * (items_size_ elements) plus the levels_ vector. Unlike a hash-table sketch this is not a table
+     * footprint; it mirrors the sketch's allocate() sites. This is the exact live heap footprint
+     * right now, not an upper bound like get_serialized_size_bytes().
+     * @return the current allocated heap size of the sketch in bytes
+     */
+    size_t get_live_bytes() const {
+      return (items_ == nullptr ? 0 : static_cast<size_t>(items_size_) * sizeof(T))
+          + levels_.capacity() * sizeof(uint32_t);
+    }
+
+    /**
      * Returns upper bound on the serialized size of a sketch given a parameter <em>k</em> and stream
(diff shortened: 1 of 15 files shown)


leerho:
Although it is possible to do this in C++, it may not be realistic in Java.  This might create a cross-language API incompatibility.
I'll have to think about this more.

leerho:
I'd like to hear from others in the community about this and how well it translates to other languages.

stojkomilos:
Sorry, I forgot to close this PR.

PR #514, opened 2026-08-20
Allow theta sketches with lg_k = 4 to match Java

SavicStefan:
Java allows theta and tuple sketches down to lg_k = 4 (16 nominal entries), but C++ rejected anything below lg_k = 5, so the two libraries disagreed on the smallest sketch you can build. 

The C++ theta implementation used a single constant, theta_constants::MIN_LG_K = 5, for two distinct roles: the floor on the user-requested nominal size (K) and the floor on the internal hash table (cache) size. Java keeps these separate ThetaUtil.MIN_LG_NOM_LONGS = 4 and MIN_LG_ARR_LONGS = 5, so Java accepts lg_k = 4 (nominal 16) while C++ rejected it in theta_base_builder::set_lg_k.

Because C++ used 5 for both, theta_base_builder::set_lg_k(4) threw, and there was no way to build a theta/tuple update sketch or union at lg_k = 4.

This only affected construction. Reading was already fine: the compact serialization format does not store lg_nom, so C++ has always been able to deserialize Java-produced lg = 4 compact sketches.

Fix, split the one constant into the two that Java has:
- MIN_LG_K -> 4, the nominal floor checked by set_lg_k.
- new MIN_LG_ARR -> 5, the internal hash-table floor used by starting_sub_multiple().

A lg_k = 4 sketch now starts with lg_cur_size (5) > lg_nom_size (4): the hash table keeps its full 32 slots and simply rebuilds down to the nominal 16, a path the existing update/rebuild logic already handles. There is no format or behavior change for lg_k >= 5.

The fix covers tuple sketches as well as theta, with no tuple-specific change. Tuple sketches have no size constants of their own their builder inherits theta_base_builder (sharing the same set_lg_k validation and starting_lg_size()), and their internal hash table is a theta_update_sketch_base (the exact table this PR resizes). So lg_k = 4 now works for tuple update sketches and unions too.

Added new min lg_k tests in both theta_sketch_test.cpp and tuple_sketch_test.cpp. Each test:
- builds at lg_k = 4 and confirms set_lg_k(MIN_LG_K - 1) throws;
- verifies the new MIN_LG_ARR floor -- after many updates the peak retained count is > 16 (2^MIN_LG_K) yet <= 32 (2^MIN_LG_ARR), i.e. the table keeps its 32 slots rather than collapsing to the nominal 16;
- confirms estimation mode, with the true count inside the 2-standard-deviation confidence bounds;
- trims down to exactly the nominal 16 entries;
- round-trips through compact serialize/deserialize.

### DIFF
--- a/theta/include/theta_constants.hpp
+++ b/theta/include/theta_constants.hpp
@@ -38,9 +38,7 @@ namespace theta_constants {
   const uint8_t MIN_LG_K = 4;
   /// max log2 of K
   const uint8_t MAX_LG_K = 26;
-  /// min log2 of the internal hash table (cache) size. The hash table needs headroom above the
-  /// nominal size to hold transient entries before rebuild, so its floor is higher than MIN_LG_K.
-  /// Matches Java's ThetaUtil.MIN_LG_ARR_LONGS.
+  /// min log2 of cache size
   const uint8_t MIN_LG_ARR = 5;
   /// default log2 of K
   const uint8_t DEFAULT_LG_K = 12;
--- a/theta/include/theta_constants.hpp
+++ b/theta/include/theta_constants.hpp
@@ -34,10 +34,14 @@ namespace theta_constants {
 
   /// max theta - signed max for compatibility with Java
   const uint64_t MAX_THETA = LLONG_MAX;
-  /// min log2 of K
-  const uint8_t MIN_LG_K = 5;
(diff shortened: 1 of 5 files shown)


vranes (approved):

stojkomilos (approved):

proost (approved):
Nice catch! Thank you!

PR #516, opened 2026-08-29
refactor: strict validation from bytes bloom filter

proost:
Following [Rust implementation](https://github.com/apache/datasketches-rust/pull/190), strict validation makes sense to me.

Although adding private function to header is not API / ABI breaking change, But i want small changes to class, so i add to new static function.

### DIFF
--- a/filters/include/bloom_filter_impl.hpp
+++ b/filters/include/bloom_filter_impl.hpp
@@ -37,6 +37,19 @@
 
 namespace datasketches {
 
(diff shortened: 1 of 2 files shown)


leerho:
I would really appreciate it if you could check the Java Implementation too.  I really want to keep Java and C++ implementations closely aligned.  Thank you.

leerho (approved):
Please also check Java and GO, I don't want them to diverge

PR #519, opened 2026-09-01
Deserialization must fail if bytes remain after the last entry

SavicStefan:
Require compact tuple sketch deserialization to consume the entire input buffer.

Adds a test for deserializing a wider summary as a narrower summary.

### DIFF
--- a/tuple/include/tuple_sketch_impl.hpp
+++ b/tuple/include/tuple_sketch_impl.hpp
@@ -590,6 +590,11 @@ compact_tuple_sketch<S, A> compact_tuple_sketch<S, A>::deserialize(const void* b
       (*summary).~S();
     }
   }
+  const size_t bytes_consumed = ptr - base;
+  if (bytes_consumed != size) {
+    throw std::out_of_range("Unexpected buffer size: bytes consumed "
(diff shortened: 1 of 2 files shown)


proost (commented):
Technically, you are right. But it can change implicit contract in the API.

SavicStefan:
> Technically, you are right. But it can change implicit contract in the API.

but like serialized format does not identify the summary type, so this is needed to detect that mismatch, wdyt?

proost:
@SavicStefan 
Now I understand your intention. you're right that the mismatch is undetectable from the format alone. But that's structural, not a gap to patch. See [this](https://github.com/apache/datasketches-java/blob/3a1bf7f2f22e78c0c227bbe409ce4b4f386f630c/src/main/java/org/apache/datasketches/tuple/SummaryDeserializer.java#L30-L39).
It can't distinguish a wrong summary type from a larger buffer or a SerDe that over-report.

And also some users will be relying on it, This change would break them.

leerho:
I agree with @proost on this.  We can't assume that a given MemorySegment is dedicated to a sketch / Summary and nothing else.  It is really up to the sketch and the user's code to figure out how many bytes to read. This is true of all our sketches that leverage MemorySegments.

PR #522, opened 2026-09-05
Finish repairing the deferred KxQ rebuild, and pin the estimator constants

leerho:
# Finish repairing the deferred KxQ rebuild, and pin the estimator constants

Depends on PR#521 (`hll-full-size-reset`); this branch is cut from it because both touch `HllUtil.hpp` and `HllArray-internal.hpp`. Held as a draft until PR#521 merges.

Three independent defects, all in the HLL estimator state.

### P1. The deferred KxQ rebuild still leaves curMin/numAtCurMin merge-order dependent

The lazy rebuild introduced in PR#364 defers recomputing KxQ and curMin after a merge. PR#512 repaired the visible damage — an empty-looking gadget discarding accumulated data, a corrupted HIP accumulator, order-dependent estimates — but not the representation the rebuild writes.

`check_rebuild_kxq_cur_min()` stores the **true minimum register value** and the **count at that minimum**. The rest of the HLL_8 code maintains a different representation, the one documented in `HllArray-internal.hpp`: `curMin` is always 0 and `numAtCurMin` is the number of zero registers, decremented by `Hll8Array::internalCouponUpdate()` as registers fill.

When the merged array has no zero registers the rebuild leaves `curMin > 0`, and from that point `numAtCurMin -= (curVal == 0)` never fires again. The stored pair freezes at whatever the rebuild computed and drifts away from the registers, so its value depends on **when** the rebuild fired — that is, on merge order.

    A = lg_k 12 HLL_4 keys [20000,30364)
    B = lg_k 10 HLL_8 keys [5000,14699)
    C = lg_k 17 HLL_4 keys [70000,72598)     // stays in SET mode
    hll_union u(7)

    order A,B,C -> curMin=6 numAtCurMin=11
    order B,C,A -> curMin=6 numAtCurMin=8
    registers identical, estimates identical (23040.86465)
    recomputed from the registers: curMin=6 numAtCurMin=8   <- A,B,C stores a wrong value 4.0.1, before #364: both orders -> curMin=0 numAtCurMin=0, byte-identical

Across 4800 randomized merge-order comparisons, 92 differed **only** in these two fields; the same sweep on the pre-#364 baseline gives 0. Estimates and bounds are unaffected, because both consumers (`getLowerBound`'s `numNonZeros` and `getHllBitMapEstimate`'s `numUnhitBuckets`) branch on `curMin == 0`. So this is a serialization-determinism defect, not an accuracy one.

`check_rebuild_kxq_cur_min()` is also missing the `curMode == HLL && tgtHllType == HLL_8` guard that datasketches-java has. On an `Hll4Array` the rebuild would overwrite `curMin` without re-encoding the nibbles, which are stored relative to it. That is unreachable today — only `Hll8Array::mergeHll` sets the flag — but `setRebuildKxqCurminFlag()` and `check_rebuild_kxq_cur_min()` are both public.

### P2. The relative-error constants are rounded literals

    HLL_HIP_RSE_FACTOR     = 0.8325546   // sqrt(ln 2)          exact: 0.8325546111576977
    HLL_NON_HIP_RSE_FACTOR = 1.03896     // sqrt(3 ln 2 - 1)    exact: 1.0389617614136892

datasketches-java computes these rather than rounding them. For `lg_k > 12` the bounds take the closed-form branch rather than the interpolation table, so the truncation is directly observable: the HIP bounds differ from Java by ~4.3e-11 relative, and the non-HIP factor is off by **1.7e-6 relative**, which skews the bounds of every union result above `lg_k` 12.

### P3. log() comes from the platform libm

`HarmonicNumbers::harmonicNumber` and `HllArray::getHllBitMapEstimate` call `std::log`. `getBitMapEstimate` computes `K * (H(K) - H(K - numHit))`, and the cancellation between two nearby harmonic numbers amplifies a 1 ULP difference in `log()` by roughly 17x. Apple libm differs from fdlibm on **0.62%** of a 400,000-sample sweep spanning subnormals to 1e300, always by 1 ULP.

datasketches-java's `StrictMath.log` is specified to be fdlibm's `__ieee754_log`, and on the tested JVM `Math.log` is bit-identical to it over 3,000,000 assorted doubles. So Java is pinned and C++ is not — and C++ is not bit-stable against itself across platforms either.

## Part 2 — The fix

- **`check_rebuild_kxq_cur_min()`** emits the canonical HLL_8 representation — `curMin = 0`,  `numAtCurMin` = number of zero registers — so the rebuilt state is indistinguishable from the incrementally-maintained state and the timing of the rebuild is not observable. Adds the HLL_8 / HLL-mode guard.
- **Full-precision RSE literals**, taken from Java's own `Double.toString()` output, with the defining expression in the comment. Literals rather than a runtime `sqrt(log(2.0))` so the value cannot vary with the platform libm.
- **`common/include/fdlibm_log.hpp`** (new): FDLIBM 5.3 `__ieee754_log`, the function `StrictMath.log` is specified to be. Copyright (C) 1993 Sun Microsystems, freely distributable with the notice preserved; the notice is in the header. Used by `harmonicNumber` and `getHllBitMapEstimate`.

  fdlibm depends on strict IEEE-754 evaluation — a fused multiply-add in the polynomial changes the result. `#pragma clang fp contract(off)` is honoured under default flags but is overridden by an explicit `-ffp-contract=fast`, which is GCC's default, so the contraction-sensitive expressions are routed through a `volatile` round-trip that the standard requires the compiler to honour.

  Verified bit-identical to Java under `-O2`, `-O3`, `-ffp-contract=on` and `-ffp-contract=fast`. `-ffast-math` breaks it, as it breaks IEEE semantics generally.

## Compatibility

Unlike PR#364 , **this PR changes serialized bytes** — that is its purpose.

- Union results: stored `curMin` / `numAtCurMin` change (to the canonical form 4.0.1 produced).
- Estimates in the linear-counting range shift by ~1e-15 relative, from the `log` change.
- Bounds at `lg_k > 12` shift by ~4.3e-11 (HIP) and ~1.7e-6 (non-HIP), from the constants.

Golden-image tests and any test asserting exact double equality against previously recorded C++ values will need updating. Reading is unaffected: images from every earlier version still deserialize, and both readers of `curMin`/`numAtCurMin` branch on `curMin == 0`, so old and new images yield identical estimates and bounds.

## Part 3 — Tests

New `hll/test/HllKxqRebuildTest.cpp` — 5 cases, 37 assertions:

- a union result is byte-identical across all six permutations of three inputs, one of which stays in SET mode so the coupon-update path into a gadget with a pending rebuild is exercised;
- reading an estimate mid-stream does not change a later `get_result()` image;
- a union result's stored `curMin`/`numAtCurMin` match a recount over its own registers;
- `get_rel_err()` matches `numStdDev * sqrt(ln 2) / sqrt(K)` and the non-HIP form at `lg_k` 13/16/21 for 1..3 standard deviations;
- `fdlibm::log` matches a nine-entry reference table on inputs where the platform libm diverges. The expected bit patterns were taken from Java's `StrictMath.log`.

Of the four assertions that also compile against the parent branch, **three fail there and pass here**. The fourth — that reading an estimate does not change a later result — already held in C++, because PR#512's eager rebuild inside `internalCouponUpdate` fires in both the peeked and unpeeked orderings. It is kept as a guard, not claimed as a regression; that observer effect is the datasketches-java defect addressed in the companion PR.

Full suite: 64 cases / 7763 assertions pass.

## Cross-language result

With the companion datasketches-java PR applied, over 1041 records — `lg_k` 4..21 x {HLL_4, HLL_6, HLL_8} x 17 sizes spanning LIST, SET and HLL modes, plus heapify round-trips and 80 deterministic pseudo-random union scenarios:

    before:  428 of 1041 records differ
    after:     0 of 1041 records differ

Zero differences in any field: both serialized forms, estimate, composite estimate, and lower and upper bounds at 1 and 2 standard deviations, compared as raw IEEE-754 bit patterns.

## Follow-ups carried in this PR

Three small changes landed here rather than as separate PRs, since each is a one- or few-line consequence of work that merged while this was open.

- **`.rat-excludes`** — `#523` added an Apache RAT license audit to CI after this branch was opened. `common/include/fdlibm_log.hpp` carries the Sun FDLIBM notice rather than an ASF header, so RAT flags it. It is excluded here and is already recorded in `LICENSE`, which is the other half of what the check requires for third-party source.

- **`HllUtil.hpp` comment** — records the date and PR in which the deprecated `FULL_SIZE_FLAG_MASK` alias was added, so a future reader can decide when it is safe to drop. A date and PR number are facts; a predicted release number is not.

- **`.gitattributes`** — `*.sk` was declared `text eol=lf`, subjecting serialized sketch images to line-ending conversion. Two of the fifteen tracked fixtures contain a CRLF pair and were reported as modified in any fresh worktree; a routine `git add` would have truncated them by a byte. Moved to the binary block. The file is `export-ignore`'d, so this cannot affect a source release.

### DIFF
--- a/hll/include/HllUtil.hpp
+++ b/hll/include/HllUtil.hpp
@@ -50,7 +50,8 @@ static const uint8_t OUT_OF_ORDER_FLAG_MASK   = 16;
 // are free.
 static const uint8_t RESERVED_FLAG_MASK_32    = 32;
 // Deprecated alias for the bit above, retained for source compatibility. Do not use.
-// (No [[deprecated]] attribute: this library targets C++11, where it is unavailable.)
+// Added 2026-09-07 (#521), first shipping in the release after 5.2.0. No [[deprecated]]:
+// this library targets C++11, so the alias warns no one.
 static const uint8_t FULL_SIZE_FLAG_MASK      = RESERVED_FLAG_MASK_32;
 
 static const uint32_t PREAMBLE_INTS_BYTE = 0;
--- a/hll/include/HllUtil.hpp
+++ b/hll/include/HllUtil.hpp
@@ -87,8 +87,8 @@ static const uint32_t EMPTY = 0;
 static const uint8_t MIN_LOG_K = 4;
 static const uint8_t MAX_LOG_K = 21;
 
-static const double HLL_HIP_RSE_FACTOR = 0.8325546; // sqrt(ln(2))
-static const double HLL_NON_HIP_RSE_FACTOR = 1.03896; // sqrt((3 * ln(2)) - 1)
+static const double HLL_HIP_RSE_FACTOR = 0.8325546111576977; // sqrt(ln(2))
+static const double HLL_NON_HIP_RSE_FACTOR = 1.0389617614136892; // sqrt((3 * ln(2)) - 1)
 static const double COUPON_RSE_FACTOR = 0.409; // at transition point not the asymptote
 static const double COUPON_RSE = COUPON_RSE_FACTOR / (1 << 13);
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,7 +11,6 @@ NOTICE text eol=lf
 *.html text eol=lf
 *.hpp text eol=lf
 *.cpp text eol=lf
-*.sk text eol=lf
 *.md text eol=lf
 *.properties text eol=lf
 *.sh text eol=lf
@@ -27,6 +26,10 @@ NOTICE text eol=lf
 *.cmd text eol=crlf
 
 # Explicitly denote all files that are truly binary and should not be modified.
+# .sk files are serialized sketch images: arbitrary bytes that must never be
+# line-ending converted. A stray CRLF pair would be rewritten to LF on check-in,
+# silently truncating the fixture by one byte.
+*.sk binary
 *.jpg binary
 *.png binary
 *.svg binary
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -7,6 +7,7 @@
 # the matching entry to LICENSE in the same commit.
 MurmurHash3\.h
 xxhash64\.h
+fdlibm_log\.hpp
 #
 # Dotfiles and repository metadata with no comment convention.
 \..*
--- a/LICENSE
+++ b/LICENSE
@@ -296,3 +296,17 @@ APPENDIX B: Additional licenses relevant to this product.
     Code Locations:
       * common/include/ceiling_power_of_2.hpp
     that is adapted from the above.
+
+    =============================================================
+    FDLIBM
+    =============================================================
+    Original source code:
+      Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+      Developed at SunSoft, a Sun Microsystems, Inc. business.
+
+      Permission to use, copy, modify, and distribute this software is freely
+      granted, provided that this notice is preserved.
+
+    Code Locations:
+      common/include/fdlibm_log.hpp
+    that is adapted from __ieee754_log in FDLIBM 5.3.
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -40,6 +40,7 @@ install(FILES
       include/conditional_back_inserter.hpp
       include/conditional_forward.hpp
       include/count_zeros.hpp
+      include/fdlibm_log.hpp
       include/inv_pow2_table.hpp
       include/kolmogorov_smirnov_impl.hpp
       include/kolmogorov_smirnov.hpp
--- a/common/include/fdlibm_log.hpp
+++ b/common/include/fdlibm_log.hpp
@@ -1,0 +1,101 @@
+// fdlibm __ieee754_log, used by Java's StrictMath.log (and by Math.log on most JVMs).
+// Derived from FDLIBM 5.3, Copyright (C) 1993 by Sun Microsystems, Inc.
+// "Permission to use, copy, modify, and distribute this software is freely granted,
+//  provided that this notice is preserved."
+#ifndef FDLIBM_LOG_HPP
+#define FDLIBM_LOG_HPP
+#include <cstdint>
+#include <cstring>
+#include <limits>
+
+namespace fdlibm {
+
+inline int32_t hi_word(double x){ uint64_t u; std::memcpy(&u,&x,8); return (int32_t)(uint32_t)(u>>32); }
+inline uint32_t lo_word(double x){ uint64_t u; std::memcpy(&u,&x,8); return (uint32_t)u; }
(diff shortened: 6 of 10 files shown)


hll/include/HllUtil.hpp
```diff
@@ -43,7 +43,12 @@ static const uint8_t FAMILY_ID = 7;
 static const uint8_t EMPTY_FLAG_MASK          = 4;
 static const uint8_t COMPACT_FLAG_MASK        = 8;
 static const uint8_t OUT_OF_ORDER_FLAG_MASK   = 16;
-static const uint8_t FULL_SIZE_FLAG_MASK      = 32;
+// Bit 32 is RESERVED and must not be written or interpreted.
+// It was formerly FULL_SIZE_FLAG_MASK here, while datasketches-java uses the same bit as
+// REBUILD_CURMIN_NUM_KXQ_MASK for its union gadget. The two meanings collided across
+// implementations, so this side no longer writes or reads it. Do not reuse: bits 64 and 128
+// are free.
+static const uint8_t RESERVED_FLAG_MASK_32    = 32;
```
proost, line 51: Like previous PR, how do you think introducing new const and deprecation instead of renaming? 

after next release, remove it.

proost (commented):
Thank you for fix.

hll/include/HllUtil.hpp
leerho: @proost, 
What you see in this diff is stale, and that's on me — #522 is stacked on the pre-fix commit of #521 and hasn't been brought up to date with master yet, so the file view still shows the old state. What you don't see is the change made in #521, which addressed this same comment. On master today:

```
static const uint8_t RESERVED_FLAG_MASK_32    = 32;
// Deprecated alias for the bit above, retained for source compatibility. Do not use.
// (No [[deprecated]] attribute: this library targets C++11, where it is unavailable.)
static const uint8_t FULL_SIZE_FLAG_MASK      = RESERVED_FLAG_MASK_32;
```

I'll merge master into this branch and it will drop out of the view.

**On removal:** with C++11 there's no [[deprecated]], so the alias warns nobody — it's purely a source-compatibility placeholder. Rather than name a version we can't predict yet, I'll record the date and PR in the comment itself so it can be traced to whatever release follows. Intent is to drop it one release after it ships. It will look something like this:

```
static const uint8_t RESERVED_FLAG_MASK_32    = 32;
// Deprecated alias for the bit above, retained for source compatibility. Do not use.
// Added 2026-09-07 (#521), first shipping in the release after 5.2.0. No [[deprecated]]:
// this library targets C++11, so the alias warns no one.
static const uint8_t FULL_SIZE_FLAG_MASK      = RESERVED_FLAG_MASK_32;
```

proost (approved):
Thank you!

PR #523, opened 2026-09-06
Add an Apache RAT license-header audit to CI

leerho:
# Add an Apache RAT license-header audit to CI

## The gap

This repository has no automated license check. Nothing in the build system knows which files are third-party, so a vendored source file can be added with **no ASF header and no LICENSE entry** and every existing check still passes.

That is not hypothetical. It happened while preparing the HLL cross-language parity work: a vendored FDLIBM header was added to `common/include/` and all 31 CI checks went green with no LICENSE entry. The omission would have surfaced only in a release audit.

## Why C++ makes this easier to get wrong than Java

In a Maven project every dependency is a coordinate in the POM, resolved and versioned by the tool. Here dependencies arrive two different ways, and only one of them is declared anywhere:

| how it arrives | where it is declared |
|---|---|
| fetched at build time — Catch2 `v2.13.9`, google-benchmark `v1.7.1` | CMake `FetchContent`, pinned by tag |
| **vendored — copied into the tree** | **`LICENSE` only, maintained by hand** |

There is no `conanfile`, no `vcpkg.json`, no manifest of any kind. The library itself links nothing external — `hll` links only `common`, both header-only INTERFACE targets — so a consumer just copies headers. That is a virtue for consumers and a hazard for compliance: the vendored files exist purely as files.

Grep is not a substitute. Of the vendored headers, `MurmurHash3.h` and `ceiling_power_of_2.hpp` state their provenance in prose without the word "Copyright", so scanning the source does not reliably enumerate them either.

## What this adds

- **`tools/rat-check.sh`** — exports the tree with `git archive` so only tracked files are examined, runs Apache RAT, parses the report and exits non-zero on any unapproved file, printing the offending paths and what to do about them.
- **`.github/workflows/rat.yml`** — runs it on push and pull request.
- **`.rat-excludes`** — the exclusions, grouped and commented, with a note at the top of the third-party group that adding a file there also requires a LICENSE entry.

Two things worth knowing about the implementation:

- **RAT always exits 0.** Version 0.16.1 has no `--fail` option, so a naive `java -jar rat.jar` step is a check that can never fail. The script parses the `N Unknown Licenses` line instead.
- **`git archive` rather than the working tree.** Only tracked files are scanned, so a local build directory cannot affect the result, and the check runs locally exactly as CI runs it. On a dirty tree RAT reports 827 unknowns; on the tracked set it reports 0.
- **RAT matches `-E` patterns against the file name, not the path.** This is undocumented in the CLI help and cost two wrong attempts; the file says so at the top.

## It found something on the first run

`hll/include/hll.private.hpp` carried no ASF header, while every sibling header in that directory does. Fixed here.

## Verification

- With the header fix, the tracked tree reports **0 unapproved files**.
- Negative test: adding a header-less file to a copy of the tree is caught — `1 Unknown Licenses`, naming the file, exit code 1.
- Exclusions were derived from what RAT actually reports on this repository, not copied from the Java project, whose list is shaped for a Maven tree (its `**/*.txt` would skip every `CMakeLists.txt`, which here does carry an ASF header and should be checked).

### DIFF
--- a/.github/workflows/rat.yml
+++ b/.github/workflows/rat.yml
@@ -1,0 +1,30 @@
+name: License Audit
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  rat:
+    name: Apache RAT
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      - name: Run Apache RAT
+        run: tools/rat-check.sh
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -1,0 +1,26 @@
+# Files excluded from the Apache RAT license-header audit.
+#
+# NOTE: RAT matches these regular expressions against the FILE NAME, not the path.
+#
+# Third-party sources vendored into this tree do not carry the ASF header; they
+# keep their own notice and are enumerated in LICENSE. When adding one here, add
+# the matching entry to LICENSE in the same commit.
+MurmurHash3\.h
+xxhash64\.h
+#
+# Dotfiles and repository metadata with no comment convention.
+\..*
+#
+# Documentation.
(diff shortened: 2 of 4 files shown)


leerho:
Please approve and merge this before PR 521

proost (approved):

PR #524, opened 2026-09-07
Add optional trim flag to update_theta_sketch::compact()

leerho:
## What changed

Adds an optional second flag to the existing `compact()` on `update_theta_sketch_alloc`:

```cpp
compact_theta_sketch_alloc<Allocator> compact(bool ordered = true, bool trim = false) const;
```

`trim = true` reduces the returned sketch to at most the nominal size `k`. The default is unchanged, and because the parameter is defaulted and `compact()` is not virtual here, every existing call site compiles untouched.

Thanks to @stojkomilos for raising the underlying need in #515. This takes a different approach — a flag on the existing method rather than a new `get_result()` — for two reasons: it avoids adding public API surface, and trimming turns out to be lossy enough that it should be an explicit opt-in rather than the natural way to obtain a result.

## Why trimming is opt-in

Relative error scales with `1 / sqrt(retained)`, so discarding entries always widens the confidence bounds, whatever mode the source is in. Measured at the default `lg_k`, comparing `compact()` against `compact(true, true)`:

|      n | retained | 2-sigma width | trimmed width | widened |
|-------:|---------:|--------------:|--------------:|--------:|
|  20000 |     5644 |        895.13 |       1120.26 |   1.25x |
|  40000 |     6037 |       1900.72 |       2419.57 |   1.27x |
|  60000 |     4800 |       3331.83 |       3585.58 |   1.08x |
| 100000 |     4285 |       5873.90 |       6024.14 |   1.03x |

The cost depends on where in the rebuild cycle the sketch is caught, which a caller cannot predict. Worst case is bounded: a sketch grown to just under the `15/16 * 2k` rebuild threshold loses nearly half its entries, widening the bounds by about `sqrt(15/8)`, roughly 37%.

Separately, a sketch in **exact mode** can retain more than `k` entries — nothing has been evicted, so theta is still 1.0. Trimming there discards real data and returns an estimating sketch:

```
n=5000  k=4096  retained=5000  source exact=YES
  compact()            -> estimate 5000     exact=YES
  compact(true, true)  -> estimate 4983.64  exact=no
```

Both effects are documented on the `trim` parameter and pinned by tests.

## Implementation

The trim path copies the retained entries first — the method is `const`, and quick select would permute the live table — then delegates to the shared `trim_to_nominal` helper described below, which partitions with `std::nth_element` at 0-based index `k` so that `entries[k]` becomes the new theta and the `k` entries below it are kept. Sorting happens only when an ordered result is requested. This is the same pivot convention already used by `rebuild()` in `theta_update_sketch_base_impl.hpp`, so trimming and rebuilding cannot drift apart.

## Review follow-up: one shared trim

@proost asked whether there was a reason not to call `shrink_to_fit` here, the way `theta_union_base::get_result` does. There was not — it was an omission. Rather than patch it, `e13672e` removes the cause: the two were copies of one algorithm and the newer copy had already drifted, so the algorithm now lives once, as `trim_to_nominal` in `theta_helpers.hpp`, and both sites call it.

| | before | after |
|---|---|---|
| `theta_union_base::get_result` | 5-line inline kernel | `theta = trim_to_nominal<EK>(entries, nominal_num, theta);` |
| `update_theta_sketch::compact` | 5-line inline kernel, **no `shrink_to_fit`** | `theta = trim_to_nominal<trivial_extract_key>(entries, nominal_size, theta);` |

The missing shrink mattered most exactly where it was missing. An update sketch retains up to `15/16 * 2k` entries between rebuilds, and the entries vector is reserved to that count before being erased down to `k`, so `std::move` handed the compact sketch the untrimmed allocation. Measured at `lg_k=12`: 7674 retained against 4096 kept, holding 1.87x the memory it reports, with 46.6% of the allocation dead. Since a bounded result is the entire point of `trim`, the caller was paying the accuracy cost documented above without getting the size back.

`theta_union_base` is instantiated by both `theta_union` and `tuple_union`, so all three paths now run identical code. Theta's `compact` is the degenerate case, where the entry is the key and `ExtractKey` is `trivial_extract_key`; tuple's entry is a `std::pair<uint64_t, Summary>` and its `pair_extract_key` recovers the hash, which is why one implementation serves both without the summary columns ever being touched.

The invariant is now stated once, in the one place it lives:

> The entry at index `nominal_size` becomes the new theta and is itself discarded, so every entry kept is strictly below the returned value. That is what keeps the estimator unbiased: theta must be an exclusive upper bound on the retained hashes. Getting this off by one does not fail loudly, it quietly biases every estimate the sketch produces.

The extraction is behaviour-preserving, verified rather than assumed: serialized bytes of both the trimmed compact result and the union result are identical before and after, across ordered and unordered at 5000, 26989 and 40000 updates. 26989 is the worst case for retained entries at this `lg_k`.

## How tested

New cases in `theta/test/theta_sketch_test.cpp`:

- **all four ordered/trim combinations**, including that the source sketch is unmodified afterwards, that trimmed output matches `trim()` + `compact()` on theta and retained set, and that every retained hash is strictly below the new theta
- **exact to estimation conversion** at n=5000
- **bounds widening** for a source already in estimation mode
- **empty and below-k** sketches, where trimming is a no-op

```
./build/theta/test/theta_test
All tests passed (20250056 assertions in 89 test cases)

./build/tuple/test/tuple_test          # shares theta_union_base
All tests passed (126786 assertions in 55 test cases)

ctest --test-dir build
100% tests passed, 0 tests failed out of 17
```

## Not included: the same flag on `update_tuple_sketch`

After this PR three of the four theta/tuple result paths can trim and one cannot:

| | trims? |
|---|---|
| `update_theta_sketch_alloc::compact(ordered, trim)` | yes, added here |
| `theta_union::get_result(ordered)` | yes, always — trimming to `k` is required of a union |
| `tuple_union::get_result(ordered)` | yes, always, via the shared `theta_union_base` |
| `update_tuple_sketch::compact(ordered)` | **no** |

The omission is deliberate rather than overlooked. Mechanically the port is small now that `trim_to_nominal` is generic over `ExtractKey`: tuple would pass `pair_extract_key<uint64_t, Summary>`, and `compact_tuple_sketch` already has the constructor the trim path needs (`tuple_sketch.hpp:650`).

What makes it a separate decision is cost, not mechanism. Tuple's `Entry` is `std::pair<uint64_t, Summary>` where theta's is a bare `uint64_t`. The trim path has to copy the retained entries before partitioning, because `compact()` is `const` and quick select permutes what it touches — so for tuple that means copying up to `15/16 * 2k` user-defined `Summary` objects of unknown expense, where theta copies 8-byte hashes. `update_tuple_sketch::compact()` is currently a one-line delegation to a constructor (`tuple_sketch_impl.hpp:257-259`) and never materialises a vector at all, so adding trim changes its cost profile. Whether to accept that copy, or offer a non-const variant that trims in place, deserves its own discussion rather than riding along here.

## Note on the Java side

`datasketches-java` should get the same capability, but it cannot use the same mechanism: `ThetaSketch.compact(boolean dstOrdered, MemorySegment dstSeg)` is `abstract` and already occupies the two-argument slot, so Java needs an overload rather than an added parameter, and `compact(true, null)` would be ambiguous. Also worth care there: `QuickSelect.selectExcludingZeros` is 1-based while `std::nth_element` is 0-based, so the Java port should assert `retained == k` the way these tests do. That work is deliberately left to a separate PR, and it inherits the same gap described above: Java's `UpdatableSketch` needs the capability as much as `UpdateSketch` does.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/theta/include/theta_helpers.hpp
+++ b/theta/include/theta_helpers.hpp
@@ -20,12 +20,44 @@
 #ifndef THETA_HELPERS_HPP_
 #define THETA_HELPERS_HPP_
 
+#include <algorithm>
+#include <cstdint>
 #include <stdexcept>
 #include <string>
+#include <vector>
 
 #include "theta_constants.hpp"
+#include "theta_comparators.hpp"
 
 namespace datasketches {
+
+/**
+ * Trims a vector of theta entries down to at most nominal_size and returns the theta the
+ * result must carry.
+ *
+ * The entry at index nominal_size becomes the new theta and is itself discarded, so every
+ * entry kept is strictly below the returned value. That is what keeps the estimator
+ * unbiased: theta must be an exclusive upper bound on the retained hashes. Getting this
+ * off by one does not fail loudly, it quietly biases every estimate the sketch produces.
+ *
+ * Capacity is released as well as size. Callers reserve an upper bound before filling, so
+ * without the shrink a trimmed result keeps the untrimmed allocation, which for an update
+ * sketch just under the rebuild threshold is nearly twice what it reports.
+ *
+ * @param entries entries to trim in place; reordered even when nothing is removed
+ * @param nominal_size the most entries to keep
+ * @param theta returned unchanged when there is nothing to trim
+ * @return the theta of the trimmed result
+ */
+template<typename ExtractKey, typename Entry, typename Allocator>
+static uint64_t trim_to_nominal(std::vector<Entry, Allocator>& entries, uint32_t nominal_size, uint64_t theta) {
+  if (entries.size() <= nominal_size) return theta;
+  std::nth_element(entries.begin(), entries.begin() + nominal_size, entries.end(), compare_by_key<ExtractKey>());
+  const uint64_t new_theta = ExtractKey()(entries[nominal_size]);
+  entries.erase(entries.begin() + nominal_size, entries.end());
+  entries.shrink_to_fit();
+  return new_theta;
+}
 
 template<typename T>
 static void check_value(T actual, T expected, const char* description) {
--- a/theta/include/theta_sketch.hpp
+++ b/theta/include/theta_sketch.hpp
@@ -330,11 +330,25 @@ public:
   void reset();
 
   /**
-   * Converts this sketch to a compact sketch (ordered or unordered).
+   * Converts this sketch to a compact sketch (ordered or unordered, trimmed or not trimmed).
+   * This does not modify the source update sketch.
    * @param ordered optional flag to specify if an ordered sketch should be produced
+   * @param trim optional flag to reduce the size of the returned sketch to at most
+   * the nominal size k, if required. An update sketch retains more than k entries
+   * between rebuilds; those extra entries below theta improve the estimate, so the
+   * default is to keep them.
+   * Trimming is lossy and is never required for correctness. It discards retained
+   * entries, and since the relative error scales with 1 / sqrt(retained), it always
+   * degrades accuracy and widens the confidence bounds, whatever mode the source is
+   * in. Worst case, a sketch grown to just under the rebuild threshold of 15/16 * 2k
+   * loses nearly half its entries, widening the bounds by about sqrt(15/8), or
+   * roughly 37%. A sketch in exact mode that retains more than k entries loses
+   * exactness as well: it is returned in estimation mode, so get_estimate() carries
+   * error where it would otherwise have returned an exact count.
+   * Only pass true if a bounded result size matters more than that accuracy.
    * @return compact sketch
    */
-  compact_theta_sketch_alloc<Allocator> compact(bool ordered = true) const;
+  compact_theta_sketch_alloc<Allocator> compact(bool ordered = true, bool trim = false) const;
 
   virtual iterator begin();
   virtual iterator end();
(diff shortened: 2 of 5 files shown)


proost (commented):
For the double-check; is same reason not to call "shrink_to_fit"?

leerho:
No reason at all — it was an omission, and a good catch.

Rather than add the line, I removed the cause. `theta_union_base::get_result` and this new trim path were two copies of one algorithm, and the newer copy had already drifted from the older one. e13672e extracts it into a single `trim_to_nominal` in `theta_helpers.hpp` and calls it from both sites. Since `theta_union_base` is instantiated by both `theta_union` and `tuple_union`, all three paths now run identical code — theta's `compact` is just the degenerate case where the entry is the key.

It mattered most exactly where it was missing. An update sketch retains up to `15/16 * 2k` entries between rebuilds, and the entries vector is reserved to that count before being erased down to `k`, so a trimmed result kept the untrimmed allocation: at `lg_k=12`, 7674 retained against 4096 kept, holding 1.87x the memory it reports with 46.6% of the allocation dead. A bounded result is the entire point of `trim`, so the caller was paying the accuracy cost without getting the size back.

The extraction is behaviour-preserving — serialized bytes of both the trimmed compact result and the union result are identical before and after, across ordered and unordered at 5000, 26989 and 40000 updates. Details in the description, along with a note on why the same flag is not being added to `update_tuple_sketch::compact()` in this PR.

proost (approved):

PR #525, opened 2026-09-16
fix(tdigest): preserve stored min and max when merging

jaideeppyne:
## Summary

- Copy the stored `min_` / `max_` from the other digest in `tdigest::merge` before compressing the combined centroid list.
- Add a regression that merges a deserialized image whose stored min/max sit strictly outside the extreme centroid means.

## Why

`merge` previously re-derived extrema from the extreme centroid means. That is exact only when those centroids have unit weight. Deserialized images (and the reference implementation format) can have weighted tails, so min/max drifted inward and rank/quantile tail interpolation used the wrong endpoints.

Same defect as apache/datasketches-rust#277.

## Validation

- `tdigest_test` including `merge preserves deserialized min max with weighted tails` and the full `[tdigest]` suite (49 cases).


leerho (approved):
Thank you for finding and fixing this!

PR #526, opened 2026-09-17
Write the single-item flag on a compact theta sketch

leerho:
## What changed

A compact theta sketch holding exactly one entry in exact mode now sets bit 5 of the flags byte, the bit Java has always written as `SINGLEITEM_FLAG_MASK`. The flags enum gains `IS_SINGLE_ITEM`, a private `is_single_item()` mirrors Java's condition (`curCount == 1 && thetaLong == Long.MAX_VALUE`, here `entries_.size() == 1 && !is_estimation_mode()`), and both version-3 writers OR it in. The compressed form never carries a single item, so `serialize_version_4` is untouched.

Readers do not change. All four deserializers test individual bits and have always ignored this one, which is why C++ already reads Java's single-item images; it now also reads its own.

## Why

This was the last theta image that differed between C++ and Java. With `c11a5ea` (zero seed hash on an empty sketch, on master but not yet released) every other theta fixture — empty, n=10 through 1,000,000, all six compressed sizes, and non-empty-no-entries — is already byte-identical. The single-item image differed only at byte 5: `0x1A` here, `0x3A` in Java.

Before, with the fixtures regenerated from master:

```
theta_n1_cpp.sk   01 03 03 00 00 1a cc 93 e5 6b 61 ee c8 80 44 20
theta_n1_java.sk  01 03 03 00 00 3a cc 93 e5 6b 61 ee c8 80 44 20
```

After this change the two files are identical and the Java `ThetaSketchCrossLanguageTest.checkCpp` byte-for-byte comparison passes for all eight sizes. The C++ `[serde_compat]` readers pass against the current Java fixtures in the other direction.

## Tests

Three new cases pin the exact serialized images, which nothing asserted before:

- the 8-byte empty image `{1, 3, 3, 0, 0, 0x1E, 0, 0}` from `serialize()`, the stream form, and `serialize_compressed()`, using a non-default seed to show the seed hash is not written;
- the 16-byte single-item image with flags `0x3A`, the seed hash present, and the one hash, from all three writers, plus a round trip through `deserialize()` both with and without the flag to show the reader does not depend on it;
- one entry in estimation mode, which has a 3-long preamble and must not get the flag.

Full theta and tuple suites pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


leerho:
The goal here is to move toward the deterministic sketches actually compare byte-for-byte across languages.  By enforcing compatibility on the _write side_ and being lenient on the _read-side_ we can maintain compatibility with older sketch images and achieve byte compatibility in our cross-language testing, which will increase our confidence in these tests.

proost (approved):
Thank you!

tisonkun (approved):

PR #527, opened 2026-09-20
frequent_items_sketch: fix is_empty() after purge clears all counters

FluorineDog:
### Problem (AI generated)

`frequent_items_sketch::is_empty()` is defined as `map.get_num_active() == 0`, conflating "no retained items" with "no observations".

A purge subtracts the median of all counters and keeps only positive ones. When every counter equals the median — typical for streams with many low-frequency items — the purge wipes **all** counters, leaving `num_active == 0` while `offset` and `total_weight` remain non-zero. Such a state is then misreported as empty, which corrupts two operations:

- `serialize()` emits the empty short format, dropping `total_weight` and `offset` on the wire. After a round-trip the sketch reads `total_weight == 0` and `get_maximum_error() == 0` — falsely looking like a fresh, exact sketch.
- `merge()` early-returns on `other.is_empty()`, silently discarding the other sketch's `total_weight` and `offset`.

### Minimal reproducer

See unittest in this PR. After serde, data is lost.

### Fix

Use the following to implement is_empty instead.

```cpp
bool is_empty() const { return total_weight == 0; }
```


leerho (approved):
This is good but not complete. I will submit a PR to fix some other related issues after this is merged.

